PR #79, opened 2022-01-21
Safe integration

Schwartz10:


Schwartz10:
Fix tests

Add more safe integration code

PR #82, opened 2022-01-21
Transaction history tweaks and detail page

### LINKED ISSUE #63: Calculate Confidence
Due to the distributed nature of the blockchain, we can never be "sure" a specific transaction has made it into a particular "block" in the blockchain. We can only be low, medium, or high confidence (generally speaking).

The way we calculate "confidence" is by how many blocks have passed since the transaction has "made it into a block".

To get the latest `height` of the blockchain (`ChainHead`), we can set up a GraphQL subscription: 

```
subscription{
  chainHead{
    height
  }
}
```

image

To calculate the confidence of a transaction, we should simply subtract: `ChainHead - TxHeight`, and use ranges to determine the confidence.

```
0-20: low confidence
20-60: medium confidence
60+ high confidence
```

^^^ These ranges should be configurable. We'll have to experiment.

For example, 
- if the ChainHead subscription returns `100`
- The transaction's height is `95`
- The confidence # is `5`
- 5 is between 0-20, this is low confidence.

Will update this ticket with the graphQL subscription information when its deployed

Relevant part of design:

image

navFooh:
- Applied some styling updates for Button/V2
- Removed `<div style={{ display: 'flex', justifyContent: 'center' }}>` wrapper for stories, was preventing divs from using available width
- Added detail page styling
- Extracted "Badge" into separate component
- Created some utility functions to be used across message table / details

This probably means that the proposal components can be updated as well using the shared utilities / components


navFooh:
@Schwartz10 Tests are still failing, but not sure if this is due to anything I changed

src/components/Button/V2.jsx
```diff
@@ -4,43 +4,31 @@ import PropTypes from 'prop-types'
...
-  text-align: center;
-  display: flex;
-  align-items: center;
-  justify-content: center;
   cursor: pointer;
-
+  display: inline-flex;
+  align-items: center;
+  gap: 0.75rem;
+  padding: 0.5em 0.75em;
```
Schwartz10, line 12: I think removing the `text-align` property made the button use in other places weird looking?

From the wallet landing screen:

image

Schwartz10:
I see one weird thing on the wallet landing screen is that the footer for some reason can now overlap the things above it:

See now:

image


See before:

image

I can't see in the code what would've caused that (or if this PR is actually the culprit...) Any ideas on that one?

src/components/Button/V2.jsx
IntentDevThijs: Good call! Hadn't noticed that button V2 was used in this context, will correct this

PR #95, opened 2022-01-25
Fix graphql codegen for msigpending query

navFooh:


navFooh:
I see now this is linked to: https://github.com/glifio/graph/issues/32

More fixes are necessary in `src/components/HistoryTables/MsigProposals/Proposal.tsx` to support the change from string to Address type. Thought to do a quick fix but will get back to it tomorrow if necessary

Schwartz10:
This is all fixed in https://github.com/glifio/react-components/pull/99

PR #135, opened 2022-02-04
Design updates

### LINKED ISSUE #93: Should Action Buttons in Head and ProposalHead have a different color theme?
@navFooh what do you think?

Green means go (speed up, accept)
Red means reject (cancel, reject)

?

Not my area of exerptise

### LINKED ISSUE #113: Action buttons should pass `small` prop and should have better hover design
Personally I like the idea of of the "Speed up" button hovering to green border and font colors and the "cancel" button hovering to red border and font colors (colors should apply to the svgs too)

Also I think we can just pass a `small` prop to each of these buttons to make them look nicer

### LINKED ISSUE #133: ButtonV2 should have disabled states

navFooh:


Schwartz10:
@navFooh can you add the same button coloring to the single `Proposal` view in this PR?

navFooh:
> @navFooh can you add the same button coloring to the single `Proposal` view in this PR?

In `ProposalHead`? That's already done

PR #192, opened 2022-03-03
Design updates / AppHeader improvements

navFooh:
Contrary to what the branch name suggests, this doesn't yet tackly the AppTile.

The setup in AppTile is a bit convoluted with passing both `large` and `small` props which have varying CSS styles throughout all the containing elements. Apart from that it seems contradictory that the props `large` and `small` can be both `true` or `false` at the same time, they also influence hover and sizing behaviour that is very hard to deduce right now. Conclusively, I couldn't find a quick fix and propose to tackle this at a later point in time.

One of the problems here that seems to reoccur throughout the codebase, is that a lot of size and layout behaviour is too much predefined on component level. This makes it hard, if not impossible, to have consistent styling and layout throughout the Apps. We're slowly starting to get more reusable layout components and stripping definitions on component level, but there is still quite a way to go.

This PR also extends the AppHeader to accomodate for an App specific icon and SubHeader.


src/components/AccountCardAlt/index.jsx
```diff
@@ -51,77 +50,75 @@ const AccountCardAlt = ({
...
+          index={index}
+          legacy={legacy}
+          path={path}
+          nDefaultWallets={nDefaultWallets}
+        />
+      </Box>
+      <Box display='flex' flexDirection='row' justifyContent='center'>
+        <StyledATag
+          target='_blank'
+          href={`https://filfox.info/en/address/${address}`}
```
Schwartz10, line 81: Should be explorer link

Schwartz10 (changes requested):
1. Take app header out of the landing page, keep phishing banner
2. Take the word "safe" out of app header (make glif name from app header optional)
3.  In "connect" pages, remove buttons from app header (just keep glif icon on top left)
4. App tile addresses should use `process.env.NEXT_PUBLIC_EXPLORER_URL` to point to the explorer for addresses
5. Make safe navigation top row (inside the safe)
- Move wallet address, safe address, and network name to the  top left (where the glif name is currently)
- Move subheader navigation to top header navigation (keep logout), and remove the top level navigation buttons (blog, wallet, explorer)
- on smaller viewports, just break things onto next lines (everything should be visible)
- Keep sticky header
- Remove line underneath the header
6. Message overview on safe has a lot of room beneath the header
7. Message history tab should send user back to /history (and should not persist the query params)
8. Proposal method badge should be tighter
image
9. Wallet - Add app header with:
- (top left) Glif logo, network name
- (top right) Send, Switch Accounts, Logout
10. Move account card and balance card to the center of the screen, side by side

PR #199, opened 2022-03-04
Adjust columns layout

navFooh:
This is a minor adjustment to the columns layout. It prevents having double margins between content that don't collapse due to flexbox


navFooh:
Reverted the fix I implemented yesterday for having the appheader sticky and cover the content underneath, while respecting the flexbox gap.

This just uses a negtive margin on the AppHeader itself allows keeping the gap between other elements, plus it's more clean

PR #230, opened 2022-03-16
Fix inspecting address checks

### LINKED ISSUE #223: Pending transaction invalidly handles inspecting address
image

navFooh:
The only thing that was really necessary was to remove the exclamation mark from:
```
disableLink={!fromAddressIsInspecting}
```

But I thought it might be better / nicer to use the `isAddrEqual` utility for comparing the addresses


Schwartz10 (approved):
I haven't tested this when no ID address exists but `"".slice(1)` is safe in node

PR #243, opened 2022-03-31
Updates for the Node app

navFooh:
Ready to merge


src/components/Icons/index.tsx
```diff
@@ -1104,3 +1090,21 @@ export const IconSearch = forwardRef<HTMLOrSVGElement, any>((props, ref) => (
     />
   </IconBase>
 ))
+
+export const IconWarn = () => (
```
Schwartz10, line 1094: Was there a reason we changed the implementation of this to move away from `IconBase` and not allow passing of props?

Schwartz10 (commented):
This looks fine for me, just wondering about the change in implementation of IconWarn

src/components/Icons/index.tsx
navFooh: I didn't notice initially that the icon was already there under the name IconCaution when I added IconWarn. Since IconWarn was implemented in the reusable WarningBox, IconCaution became redundant.

I'm fine adding it as an extension of IconBase and allow to pass props if you want. Although usually in favor leaner implementations where possible. If we choose styled-system as our preferred styling method, I'm not sure of we need to have the option of passing various props.

Schwartz10: Understood - are we sure IconCaution is not used in any other app? If not, i'll merge this

navFooh: I didn't find any other occurences, should be fine!

PR #260, opened 2022-04-13
Tx replacement updates

navFooh:
- Adds hooks for getting lotus message and gas parameters for tx replacement
- Adds `ButtonV2Link` which is basically a `SmartLink` with `ButtonV2` styling. This greatly simplifies cases where links look like buttons, removing the need for `router.navigate()` or `window.location.assign()` in `onClick` handlers
- `speedUpHref` and `cancelHref` can always be a string because the CID is already known when passing the props
- Adds `InputV2` components which allow for more consistent and easier styling of forms
- Extends functionality of layout components


src/components/Button/V2.jsx
```diff
@@ -2,145 +2,125 @@ import styled, { css } from 'styled-components'
...
+          background: var(--white);
+          svg path {
+            stroke: var(--red-medium);
+          }
+        }
+      `}
+
+      ${props.disabled &&
+      css`
+        pointer-events: none;
```
Schwartz10, line 109: Let's add `cursor: disabled;` here ?

src/components/InputV2/BigInt.tsx
```diff
@@ -0,0 +1,96 @@
...
+  onChange: (value: BigInt) => void
+  onBlur: (value: BigInt) => void
+  setHasError: (hasError: boolean) => void
+} & Omit<TextInputProps, 'error' | 'type' | 'value' | 'onChange' | 'onBlur'>
+
+// "onChange" and "onBlur" remain of type "PropTypes.func"
+// eslint-disable-next-line
+const { error, type, value, ...bigIntProps } = TextInput.propTypes
+
+// @types/prop-types is outdated
```
Schwartz10, line 101: Instead of ignoring type errors, should we create a custom PropType for `bigint`? Here's an example:

https://github.com/glifio/react-components/blob/primary/src/customPropTypes.js#L18-L32

The only problem with the example above is that I'm not sure the chaining of `.isRequired` still works... Would have to double check.

This isn't the biggest deal either way.

src/components/InputV2/BigInt.tsx
```diff
@@ -0,0 +1,96 @@
+import PropTypes from 'prop-types'
+import { useState } from 'react'
+import { TextInput, TextInputProps } from './Text'
+
+const isBigInt = value => typeof value === 'bigint'
```
Schwartz10: does  `typeof <FilecoinNumber> === bigint`? If not, I think we should figure out how to support FilecoinNumber.

src/components/InputV2/BigInt.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * type: always "number" for bigint input
+ * value: needs to be of type "BigInt" / "PropTypes.bigint"
+ * onChange: needs to take "BigInt" type argument
+ * onBlur: needs to take "BigInt" type argument
+ */
+
+export type BigIntInputProps = {
+  min?: BigInt
+  max?: BigInt
+  value?: BigInt
```
Schwartz10: Why is `value` not required but `onChange` and `onBlur` are?

src/components/InputV2/BigInt.tsx
```diff
@@ -0,0 +1,96 @@
...
+  onChange: (value: BigInt) => void
+  onBlur: (value: BigInt) => void
```
Schwartz10: Do we definitely want to make this type required?

Schwartz10: Ah I see you provided default props for this. I would still mark these as optional `?` because ts will complain on the consuming app's side, it doesn't know about the defaultProps (afaik)

src/components/InputV2/Number.tsx
```diff
@@ -0,0 +1,84 @@
...
+      setError(`Cannot be more than ${max}`)
+      setHasError(true)
+    }
+    onBlur(number)
+  }
+
+  return (
+    <TextInput
+      error={error}
+      type='number'
```
Schwartz10: Small nit - `TextInput` that takes a `type number` is interesting. We could follow the existing pattern and use some type of `BaseInput`?

src/components/InputV2/Number.tsx
```diff
@@ -0,0 +1,84 @@
...
+  max: PropTypes.number,
+  value: PropTypes.number,
+  setHasError: PropTypes.func,
+  ...numberProps
+}
+
+NumberInput.defaultProps = {
+  min: -Infinity,
+  max: Infinity,
+  value: NaN,
```
Schwartz10, line 109: Interesting defaultProp decision - why did you chose this value?

src/components/InputV2/Text.tsx
```diff
@@ -0,0 +1,101 @@
...
+  type: string
+  autofocus: boolean
+  disabled: boolean
+  placeholder: string
+  controlled: boolean
+  value: string
+  unit: string
+  onChange: (value: string) => void
+  onBlur: (value: string) => void
+}
```
Schwartz10: For the sake of typescript and consuming apps, should we mark all of these as optional? Are any _actually_ required?

Schwartz10 (changes requested):
Great work on this. Some thoughts:

1. Inputs

With the input elements, I'm still questioning the following decisions:
- How to handle errors. In certain scenarios (like the Send page in the Wallet), it's helpful for the parent component to know if the input component has an error, and what error occurred in the input. For example, it's helpful to know if a user entered a bad value in an input, so we disable the "next" button (and do not allow a user to continue progressing through a transaction if they have an invalid form field).
- using js `bigint` types - i don't know much about them, so just voicing that I need to read up a bit more to get a better understanding of what this look like on the consuming app side. Is the dev experience needlessly complicated when creating a simple input?

2. Toggle

Right now the toggle button looks great, but I think we should have an "active" and "inactive" state. Right now it just looks like "active" and "active". To differentiate between "inactive" and "disabled", we could make the `cursor: pointer` vs `cursor: disabled`?

3. Home for `useGetMessage` and `useGetReplaceMessageGasParams`

I think these are better suited to go in the `wallet-provider-react` repo. We don't have to add any deps to this repo and these to me are more "wallet operation" esque things, rather than UI react components. This line is blurry, but I think it generally makes more sense to me to do it this way.

src/components/Button/V2.jsx
navFooh: Do you mean `cursor: none;` or `cursor: default;`? `disabled` is not a valid `cursor` value.

In any case, `pointer-events: none;` causes the cursor to completely ignore this element, so any click, hovers or cursor states are never invoked. So no `cursor` property is going to have any effect here.

src/components/InputV2/BigInt.tsx
navFooh: Yeah the annoying thing is that the PropType does actually exists, but the typing information doesn't. I assume this will be fixed when [@types/prop-types](https://www.npmjs.com/package/@types/prop-types) is bumped to 15.8.x

src/components/InputV2/BigInt.tsx
navFooh: That's a great idea, I think supporting FilecoinNumber as a prop makes a lot of sense!

src/components/InputV2/BigInt.tsx
navFooh: That's a good call, I'll check it out

src/components/InputV2/BigInt.tsx
navFooh: Neither are required but `onChange` and `onBlur` can be provided as default values with a noop function. For a BigInt value this makes less sense I think, but I understand it looks confusing. I'll have another look at it

src/components/InputV2/Number.tsx
navFooh: I see how that's semantically counterintuitive :)

The idea for TextInput was that it is the base for anything that looks like a html text input field, including the types `text`, `email`, `password`, `search` and `number`. But I see the merit of calling it `BaseInput` and implementing a separate `TextInput`, `PasswordInput`, etc

src/components/InputV2/Number.tsx
navFooh: The `min` / `max` values of `Infinity` are convenient so it is always safe to check if `value` is bigger or smaller, even if they're not provided.

For `value` I set it to `NaN` so we can also differentiate between omitting `value` or passing `0`. When `value` is omitted, the input field is empty. When the app passes `0`, it shows `0`. This could also be achieved by omitting the default and checking for `undefined`, but I think this is more explicit.

src/components/InputV2/Text.tsx
navFooh: They're all passed as default props, so all of them will actually contain valid values.

I think TypeScript / VSCode are smart enough to understand this and won't complain if the apps don't pass them.

I some cases I do actually prefer having a default, so it is always safe to call `onChange()` instead of going `onChange && onChange()` for example

navFooh:
> * How to handle errors. In certain scenarios (like the Send page in the Wallet), it's helpful for the parent component to know if the input component has an error, and what error occurred in the input. For example, it's helpful to know if a user entered a bad value in an input, so we disable the "next" button (and do not allow a user to continue progressing through a transaction if they have an invalid form field).

For number / Filecoin / BigInt inputs, there is a property called `setHasError(hasError: boolean)` that will alert the parent of faulty input, this is because the rules for validation are simple (bigger than min and smaller than max). It's only a boolean so it doesn't relay the error message, but the input takes care of displaying that itself. It does allow us to disable Send buttons when there are any errors.

For textual inputs, the app can validate the value and set an error on the input field. I'd have to see more use cases to check if we can generalize this further, but I think it'd be easy to do this as we roll it out.

> * using js `bigint` types - i don't know much about them, so just voicing that I need to read up a bit more to get a better understanding of what this look like on the consuming app side. Is the dev experience needlessly complicated when creating a simple input?

I don't think so, a `BigInt` can be created from a `number` or `string`, but as we mostly deal with `Filecoin` objects, it might be better to use those instead of `BigInt`. I'll look into it.

> 2. Toggle
> 
> Right now the toggle button looks great, but I think we should have an "active" and "inactive" state. Right now it just looks like "active" and "active". To differentiate between "inactive" and "disabled", we could make the `cursor: pointer` vs `cursor: disabled`?

I think what is mostly confusing is that the toggle colors for `hover` / `focus` are the same as `active`. When the toggle is disabled it is gray and doesn't receive any pointer events. I'll look into a more contrasting color for the active state.

> 3. Home for `useGetMessage` and `useGetReplaceMessageGasParams`
> 
> I think these are better suited to go in the `wallet-provider-react` repo. We don't have to add any deps to this repo and these to me are more "wallet operation" esque things, rather than UI react components. This line is blurry, but I think it generally makes more sense to me to do it this way.

I was kind of expecting this one, but I agree that the line can be a bit blurry between `react-components` and `wallet-provider-react` functionality. I'm still wondering a bit what the motivation is to separate the two, but I'm fine with moving them to `wallet-provider-react`.

navFooh:
I think for the toggle, we might remove the focus / hover states altogether. What do you think of these states:

Disabled:
image

Inactive:
image

Active:
image

src/components/InputV2/BigInt.tsx
Schwartz10: Hey @navFooh let's mark these as optional still in the typescript types?

src/components/InputV2/Base.tsx
```diff
@@ -0,0 +1,101 @@
...
+        onBlur={e => onBlur(e.target.value)}
+        style={{ paddingRight: `${1 + 0.75 * unit.length}em` }}
+      />
+      {unit && <span className='unit'>{unit}</span>}
+    </div>
+    {vertical && error && <span className='error'>{error}</span>}
+  </Label>
+)
+
+export interface BaseInputProps {
```
Schwartz10, line 52: Hey I would make this a `Partial<` type, so anything rendering it doesn't get into typing trouble by not supplying a certain prop (same thing with default props not being considered by typescript compiler)

src/components/InputV2/Filecoin.tsx
```diff
@@ -0,0 +1,159 @@
+import PropTypes from 'prop-types'
+import { useState } from 'react'
+import { FilecoinNumber } from '@glif/filecoin-number'
+import { BaseInput, BaseInputProps } from './Base'
+import { FILECOIN_NUMBER_PROP } from '../../customPropTypes'
+
+type FilecoinDenomination = 'fil' | 'picofil' | 'attofil'
```
Schwartz10, line 7: Would be nice to pull this from the `FilecoinNumber` package itself, but it's not exported... grrr https://github.com/glifio/modules/blob/primary/packages/filecoin-number/src/FilecoinNumber.ts#L7

src/components/InputV2/Filecoin.tsx
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+export type FilecoinInputProps = {
+  min?: FilecoinNumber
+  max?: FilecoinNumber
+  value?: FilecoinNumber
+  denom: FilecoinDenomination
+  onChange: (value: FilecoinNumber) => void
+  onBlur: (value: FilecoinNumber) => void
+  setHasError: (hasError: boolean) => void
```
Schwartz10, line 135: Let's mark these as optional so typescript doesn't complain?

src/components/InputV2/BigInt.tsx
Schwartz10: Oops missed the new `Omit`

src/components/InputV2/Filecoin.tsx
```diff
@@ -0,0 +1,159 @@
...
+ * value: needs to be of type "FilecoinNumber" / "FILECOIN_NUMBER_PROP"
+ * onChange: needs to take "FilecoinNumber" type argument
+ * onBlur: needs to take "FilecoinNumber" type argument
+ */
+
+export type FilecoinInputProps = {
+  min?: FilecoinNumber
+  max?: FilecoinNumber
+  value?: FilecoinNumber
+  denom: FilecoinDenomination
```
Schwartz10: if we're supplying a default prop for it, let's add this to the `Omit`?

src/components/InputV2/Text.tsx
Schwartz10: Agree with the decision to pass props for the functions, but are you sure typescript doesn't complain given certain configurations? I understand the linter might be smart enough

Schwartz10:
> I'm still wondering a bit what the motivation is to separate the two (react-components and wallet-provider-react), but I'm fine with moving them to wallet-provider-react.

Short answer - not all apps (explorer, glif home) require "wallet" things

Schwartz10 (approved):

PR #262, opened 2022-04-14
Make custom proptypes requirable

navFooh:
I've tried the approach of creating proptypes that can be used as such:
```
FILECOIN_NUMBER_PROPTYPE
FILECOIN_NUMBER_PROPTYPE.isRequired
```
But didn't get it to work without TypeScript complaining about the `isRequired` variants.
So ended up with this:
```
FILECOIN_NUMBER_PROPTYPE
FILECOIN_NUMBER_PROPTYPE_REQUIRED
``` 
Another thing I noticed is that each app redefines these proptypes, I imagine they could be shared from `react-components`, so I've added them to the export of `src\index.ts`


navFooh:
Also removed the `null` returns from custom proptypes, they shouldn't be necessary. And updated the logic for `FILECOIN_NUMBER_PROPTYPE`, since the `representsANum` check didn't seem to make sense.

PR #271, opened 2022-04-22
Layout updates

navFooh:
- Remove replaced messages from transaction history
- Add error icon on error boxes
- Add Address input
- Add Params input
- Set default minimum for Filecoin input to 0 FIL
- Add story for Boxes
- Fixed some styling
- Fixed some comments


src/components/HistoryTables/useAllMessages/index.ts
```diff
@@ -157,9 +157,21 @@ export const useAllMessages = (address: string, _offset: number = 0) => {
...
-        .filter(msg => !confirmedCids.has(msg.cid))
-        .sort((a, b) => Number(b.nonce) - Number(a.nonce))
+      return (
+        pendingMsgList
+          // Remove replaced messages
+          .filter((msg, i, arr) => {
+            return !arr.find(
+              m =>
+                m.nonce === msg.nonce &&
+                new BigNumber(m.gasPremium).isGreaterThan(msg.gasPremium)
```
Schwartz10: Two notes:

1. Let's move this logic into the `usePendingMessages` hook
2. We need to compare `from` and `nonce` - since an address sent _to_ this address could have the same nonce as once that was sent _from_ this address, and that's not a replaced transaction

src/components/InputV2/Address.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const onBlurBase = () => {
+    setHasFocus(false)
+    onBlur()
+  }
+
+  useEffect(() => {
+    const isValid = validateAddressString(value)
+    setTruncated(isValid ? truncateAddress(value) : value)
+    setError(isValid ? '' : 'Needs to be a valid address')
+    setIsValid(isValid)
```
Schwartz10: I'm confused a bit by how this doesn't trigger an infinite loop? I feel like we could:

- remove the stateful `truncated` variable, since this can be computed on the fly (or at least in a `useMemo` 
- Only `setState` inside the `useEffect` when we need to (if isValid status has changed and if an error state should change)

Schwartz10: I see this pattern is used elsewhere, and wondering if it matters or if react under the hood got smart about this type of thing (setting state to an existing value where the value hasn't changed, doesn't actually trigger a rerender).

https://github.com/glifio/react-components/blob/21d449205ca7c2e9da561d15397d645ed072ae04/src/components/InputV2/Filecoin.tsx#L97-L98 i see it used here too, but missed flagging it

Swear there used to be some linter warning for this

src/components/InputV2/Params.tsx
```diff
@@ -0,0 +1,95 @@
...
+      const buffer = Buffer.from(value, 'base64')
+      const result = buffer.toString('base64')
+      if (result !== value) throw new Error()
+    } catch (e) {
+      setError('Needs to be valid Base64')
+      setIsValid(false)
+      return
+    }
+    setError('')
+    setIsValid(true)
```
Schwartz10: Same Q's here

Schwartz10 (commented):
Mainly wondering about the questionable infinite looping scenarios

src/components/InputV2/Address.tsx
navFooh: This useEffect is only executed when the values in the supplied array change, which we are not setting the state of. So either way I don't think this could cause an infinite loop. But I think you're right that we maybe should use useMemo for the truncated version

src/components/HistoryTables/useAllMessages/index.ts
navFooh: Good call about the `from` comparison!

navFooh: Isn't `usePendingMessages` auto-generated by graphql codegen? Can't browse easily since I'm not at my pc right now

Schwartz10: https://github.com/glifio/react-components/blob/primary/src/components/HistoryTables/useAllMessages/index.ts#L22 the pending messages logic is a bit more complex, you could actually implement this in `uniqueifyMessages` https://github.com/glifio/react-components/blob/primary/src/utils/uniqueifyMsgs/index.ts

src/components/InputV2/Params.tsx
navFooh: It applies to all input with validation, so let's keep the conversation above to decide on further action

src/components/InputV2/Address.tsx
navFooh: I've rewritten the validated inputs and removed the local `setState` calls from `useEffect`.

We only still call `setIsValid` from `useEffect` in order to set the `isValid` state on the parent component. I don't see any way around this. I will only cause issues when the parent component modifies `value` based on `isValid`, which could cause the validator to re-run and cause an infinite loop. But I think we can make it a rule that this should never happen, I don't see a situation where that makes sense.

src/components/HistoryTables/useAllMessages/index.ts
navFooh: moved!

PR #280, opened 2022-04-29
Tx updates

navFooh:


src/components/Transaction/Confirm.tsx
```diff
@@ -0,0 +1,130 @@
...
+  msig,
+  method,
+  approvalsLeft
+}: TransactionConfirmProps) => {
+  const action = getAction(msig, method)
+  const resolution = getResolution(loginOption, msig)
+  return (
+    <InfoBox>
+      <p
+        dangerouslySetInnerHTML={{
```
Schwartz10: Why dangerouslySetInnerHTML here?

src/components/Transaction/Confirm.tsx
```diff
@@ -0,0 +1,130 @@
...
+      return "lock this Safe's balance"
+    default:
+      return 'perform this action'
+  }
+}
+
+function getResolution(loginOption: LoginOption, msig: boolean): string {
+  switch (loginOption) {
+    case LoginOption.LEDGER:
+      return 'review the details and confirm the transaction on your <b>Ledger Device</b>.'
```
Schwartz10: This may be an artifact of bad engineering on my part, but should the `msig` check be used here too like it is in metamask?

If it's a send, we can be a little more specific

src/components/Transaction/Header.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+`
+
+export const TransactionHeader = ({
+  address,
+  balance
+}: TransactionHeaderProps) => {
+  return (
+    <Header>
```
Schwartz10, line 34: The pleasure of seeing no GlyphAcronym's here :D

src/customPropTypes.ts
```diff
@@ -59,28 +71,54 @@ export const FILECOIN_NUMBER_PROPTYPE: Requireable<any> = Object.assign(
...
+
+export enum LoginOption {
+  LEDGER = 'LEDGER',
+  METAMASK = 'METAMASK',
+  CREATE_MNEMONIC = 'CREATE_MNEMONIC',
+  IMPORT_MNEMONIC = 'IMPORT_MNEMONIC',
+  IMPORT_SINGLE_KEY = 'IMPORT_SINGLE_KEY'
+}
+
+export const LOGIN_OPTION_PROPTYPE = oneOf(
```
Schwartz10, line 118: Nice

src/utils/useGetMessage.ts
```diff
@@ -0,0 +1,40 @@
...
+  const [error, setError] = useState<Error | null>(null)
+
+  useEffect(() => {
+    setMessage(null)
+    setError(null)
+    if (validateCID(cid)) {
+      setLoading(true)
+      lotusRPC
+        .request('ChainGetMessage', { '/': cid })
+        .then((m: LotusMessage) => setMessage(m))
```
Schwartz10: Not sure if this actually matters, but `m` is not actually an `instanceof LotusMessage`. It should have the fields correctly capitalized, but it doesn't have any of the other prototype methods, which maybe end up causing weird errors upstream.

Can be solved by `Message.fromLotusType(m)` somewhere in this promise chain

src/utils/useGetReplaceMessageGasParams.ts
```diff
@@ -0,0 +1,55 @@
...
+  const [loading, setLoading] = useState<boolean>(false)
+  const [error, setError] = useState<Error | null>(null)
+
+  useEffect(() => {
+    const getGasParams = minimum
+      ? provider.getReplaceMessageMinGasParams
+      : provider.getReplaceMessageGasParams
+    setGasParams(null)
+    setError(null)
+    if (message) {
```
Schwartz10, line 48: I think here we should also check if we're not already loading.

It feels like something (looking at the `provider` dep in the useEffect dependency array) could trigger this `useEffect` hook to rerender unexpectedly, which could cause some trouble when we keep resetting gasParams in state.

Schwartz10 (commented):
Nice work! Some small comments in here that I think are worth addressing

src/utils/useGetReplaceMessageGasParams.ts
navFooh: I wonder, if the message does change while loading and we ignore it, we might end up with old / wrong data? Maybe this doesn't apply in the current use case of tx replacement, but it doesn't feel completely right.

Do you think that resetting the gas params could cause the provider to change?

src/utils/useGetMessage.ts
navFooh: I see, yeah then maybe this should return a `Message` instead of a `LotusMessage`

src/components/Transaction/Confirm.tsx
navFooh: That was due to being able to set the `<b>` elements in strings, but probably better to return JSX, I'll update that!

src/components/Transaction/Confirm.tsx
navFooh: Yeah that was due to the reverse engineering of the previous code, will update as well!

src/utils/useGetMessage.ts
Schwartz10: I like the concept of `useGetMessage` returning a `Message` because then upstream consumers of it can do as they please with the instance of Message class.

Also, to solve this wonky typing problem (if there even is one), the `request` method takes an optional generic property, so this should work fine without ts errors (i think):

```
lotusRPC.request<LotusMessage>(...)
.then((m) => setMessage(Message.fromLotusType(m)))
```

src/utils/useGetReplaceMessageGasParams.ts
Schwartz10: I foresee a weird series of events where the `provider` could change, mostly around the way the Ledger Device wallet subprovider works. We basically have to reinstantiate a Ledger wallet subprovider instance in certain circumstances, which i think would trigger a falsey equality check by react's dependency array checker

But you're totally right - a user could edit some input field in transaction replacement while the current message's gas is being fetched (the node can sometimes stall on these calls too). 

Maybe we just do a better check on the gas params to see if they've changed since what's already in state? And if so, only set the state if any one of the values has changed?

src/utils/useGetMessage.ts
navFooh: I agree it makes more sense to pass around `Message` objects instead of `LotusMessage`. I've changed the other hooks to take `Message` as input parameter as well

src/utils/useGetReplaceMessageGasParams.ts
navFooh: Curious what you think about this:
https://github.com/glifio/react-components/pull/282

PR #284, opened 2022-05-05
Input V2 tests

### LINKED ISSUE #250: Params should wrap text
Check out params at the bottom of this message

https://explorer.glif.link/message/?cid=bafy2bzacebouc4ibqb5uadskoxln2f6xmfpvqq2v5db6ddmp4mwxrinzkip6c

### LINKED ISSUE #285: Extra "$" in Transaction component
image

navFooh:
Also added a `RequireableInput` component to add error handing for empty text inputs.
And fixed the number inputs so that it's possible to have leading / trailing zeroes during input.

### DIFF
--- a/src/components/HistoryTables/detail.tsx
+++ b/src/components/HistoryTables/detail.tsx
@@ -207,7 +207,13 @@ export const Line = ({ label, depth, children }: LineProps) => (
     <Box minWidth='200px' flex='0 1 25%'>
       {label}
     </Box>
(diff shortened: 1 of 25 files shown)


Schwartz10 (changes requested):
Will resolve https://github.com/glifio/wallet/issues/1010 

@navFooh one other small change - 

Let's change the copy [here](https://github.com/glifio/react-components/blob/primary/src/components/InputV2/Filecoin.tsx#L64) to just:  "Invalid value"

The longer text can wrap the line and it shifts the height of the form, which looks a bitch glitchy

PR #314, opened 2022-05-12
Design updates

navFooh:
- Renamed `Transaction.Header` to `Transaction.Balance`, since it better indicates what it's used for
- Added support for using `Transaction.Balance` in Safe, displaying both Signer and Safe balance
- Renamed `Transaction.State` to `Transaction.Header`, since it better indicates what it's used for
- Tweaked `Transaction.Header` so that title doesn't disappear when loading or confirming
- Added `Transaction.Header` to StoryBook
- Added `Transaction.Fee` component which takes care of a bit of the tedious logic around handling the Transaction Fee input
- Added `TxState.messageLoading` which happens when TX Replacement first loads the existing message. This allows us to also show a loader glyph for the initial load and simplifies a lot in the `Replace` component
- Implemented Expert Mode toggle for TX fee and improved TX fee behaviour


src/utils/useGetGasParams.ts
```diff
@@ -46,17 +32,16 @@ export const useGetGasParams = (
...
             gasLimit: new FilecoinNumber(m.gasLimit, 'attofil')
-          }
-          if (didUpdate(newGasParams)) setGasParams(newGasParams)
+          })
         })
         .catch((e: Error) => setError(e))
         .finally(() => setLoading(false))
     }
-  }, [provider, message, maxFee, didUpdate])
+  }, [provider, message, maxFee])
```
navFooh, line 44: @Schwartz10 This had to be reverted. Basically what it creates is a condition where we first set `gasParams` to `null` and then we never update it after because our cached values are still the same. This is actually very easy to reproduce and creates quite a problem.

The issue we were trying to solve here as I remember is to keep the values when the instance of provider changes. I'm not sure if we should worry about it too much, as it sounded like an edge case? If it does happen, we just reload the gas params once more, it doesn't seem like a biggie.

Schwartz10 (changes requested):
@navFooh this PR looks good to me, but the transaction fee component needs one change.

When a user enters the maxFee in expert mode, the max fee shown in the input UI should _be_ the max fee entered by the user. Right now, when a user enters a max fee, the new gas params are estimated, but then a new max fee is set, overwriting the inputted fee in the UI. I could imagine this being frustrating UX.

I imagine what's happening is the gas param estimates are coming back with a lower max fee than the param set in the UI. In this case, if maxFee calculated is less than maxFee entered, just show maxFee entered... It's not necessarily the most accurate, but it's an overestimate (not dangerous) and probably less confusing for the user?

navFooh:
> I imagine what's happening is the gas param estimates are coming back with a lower max fee than the param set in the UI. In this case, if maxFee calculated is less than maxFee entered, just show maxFee entered... It's not necessarily the most accurate, but it's an overestimate (not dangerous) and probably less confusing for the user?

Yeah, this is indeed what's happening. I see your point of it being potentially frustrating when your own input is overwritten by the calculation. Though I do see merit in displaying the actually calculated gas fees in the message "you will not pay more than ..." and in the total below.

What struck me as weird in the old send form, was that you could enter insane amounts for tx fees, like for example 0.5 FIL. If the amount to send was 1 FIL, in the total below it would show 1.5 FIL. However, the values that are actually sent by the transaction (which would display in the Metamask popup) would be waaay smaller, because the `getGasParams` method we call never returns higher gas fees than the default.

So then I propose to not change the input field, as you request, but still display the actual max gas fee in the total, which is how it is before this PR as well.

PR #322, opened 2022-05-18
Port over all logic from wallet-provider-react

Schwartz10:


src/customPropTypes.ts
```diff
@@ -153,3 +153,29 @@ export enum TxState {
...
+          `Invalid prop: ${propName} supplied to ${componentName}. Validation failed.`
+        )
+      }
+    }
+  }
+
+export const COIN_TYPE_PROPTYPE: Requireable<any> = Object.assign(
+  createCoinTypePropType(false),
+  { isRequired: createFilecoinNumberPropType(true) }
+)
```
navFooh: Just wondering, can't this be:

```
export const COIN_TYPE_PROPTYPE = oneOf(['f', 't'])
```

navFooh (changes requested):

src/customPropTypes.ts
Schwartz10: Yes

PR #336, opened 2022-05-19
Add delete button and timeout event to InputV2

navFooh:
- Adds a delete button next to the input when `deletable` is `true`
- Calls `onDelete` when the delete button is clicked
- Calls `onTimeout` 1s after the user stops entering text in the input


Schwartz10 (changes requested):
As discussed in discord, i think we should reconsider this approach of using a timer to set the input value because as far as i can tell, it's not precise enough, which could cause problems - sorry for potentially taking us down the wrong path!

PR #340, opened 2022-05-20
Input timeout

navFooh:
Just in case we do decide to implement this


Schwartz10:
@navFooh should we close this?

navFooh:
> @navFooh should we close this?

Yeah totally

PR #349, opened 2022-05-25
Updated TX fee flow

navFooh:


src/components/InputV2/Base.tsx
```diff
@@ -59,6 +64,7 @@ export const BaseInput = ({
           onChange={e => onChange(e.target.value)}
           onFocus={() => onFocus()}
           onBlur={() => onBlur()}
+          onClick={() => onClick()}
```
Schwartz10, line 67: Small nit, that may get typescript compiler angry with us... 

instead of passing anonymous functions to these props (which get reinitialized on every rerender i believe?), could we just:

```
          onFocus={onFocus}
          onBlur={onBlur}
          onClick={onClick}
```

If typescript gets angry about this, just leave it

src/components/Transaction/Form.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const [inputFee, setInputFee] = useState<FilecoinNumber | null>(null)
+
+  // Calculate transaction fee (fee cap times limit)
+  useEffect(() => {
+    if (!messageWithGas) return setTxFee(null)
+    const feeCapStr = messageWithGas.gasFeeCap.toFixed(0, BigNumber.ROUND_CEIL)
+    const feeCap = new FilecoinNumber(feeCapStr, 'attofil')
+    const limit = messageWithGas.gasLimit
+    setTxFee(getMaxGasFee(feeCap, limit))
+  }, [setTxFee, messageWithGas])
```
Schwartz10, line 63: Is there a reason we have to `useEffect` here instead of `useMemo` and pass `messageWithGas` in the dep array?

src/components/Transaction/Form.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  //Load gas params with the current message and input fee
+  const getGasParams = async () => {
+    setTxState(TxState.LoadingTxFee)
+    setGasParamsError(null)
+    setMessageWithGas(null)
+    try {
+      const provider = await getProvider()
```
Schwartz10, line 71: This is counter intuitive, and the result of bad interface design on the `filecoin-wallet-provider` package, but we don't have to `getProvider` here, and instead, we can just import the `walletProvider` from the `useWalletProvider` [hook](https://github.com/glifio/react-components/pull/349/files#diff-2dc6a1869aaab4696f200f79e928925859087c898950dd8eaf93a36ca5c0a056R40).

`gasEstimateMessageGas` would be better off as a static method on the filecoin-wallet-provider class rather than an instance method, since it doesn't _really_ need a wallet "subprovider" in order to complete this call.

using `getProvider` here might not hurt, but it does a lot of unnecessary work (like connecting to the ledger device)

src/components/Transaction/Form.tsx
```diff
@@ -0,0 +1,199 @@
...
+      const lotusMessage = newMessage.toLotusType()
+      const signedMessage = await provider.wallet.sign(
+        wallet.address,
+        lotusMessage
+      )
+      setTxState(TxState.MPoolPushing)
+      const msgValid = await provider.simulateMessage(lotusMessage)
+      if (!msgValid) {
+        throw new Error('Filecoin message invalid. No gas or fees were spent.')
+      }
```
Schwartz10, line 114: Let's actually move this bit above the `AwaitingConfirmation` set state action - this way we can know if the message is invalid before we ask the user to sign it.

src/components/Transaction/Form.tsx
```diff
@@ -0,0 +1,199 @@
...
+    try {
+      const provider = await getProvider()
+      const nonce = await provider.getNonce(wallet.address)
+      const newMessage = new Message({
+        to: messageWithGas.to,
+        from: messageWithGas.from,
+        nonce,
+        value: messageWithGas.value,
+        method: messageWithGas.method,
+        params: getParams ? getParams(nonce) : messageWithGas.params,
```
Schwartz10, line 99: is this `getParams` helper necessary? I'm guessing this might have something to do with the wasm bits in the safe or transaction replacement? What's the use case here, and why is the nonce important in determining it?

I'd vote to try and get rid of that prop

src/components/Transaction/Form.tsx
```diff
@@ -0,0 +1,199 @@
...
+        {total && <TransactionTotal total={total} />}
+      </ShadowBox>
+      <TransactionButtons
+        backDisabled={
+          txState !== TxState.FillingForm && txState !== TxState.FillingTxFee
+        }
+        nextDisabled={
+          (txState !== TxState.FillingForm || !message) &&
+          (txState !== TxState.FillingTxFee || !messageWithGas)
+        }
```
Schwartz10, line 162: Wondering if we need to disable the next button if:

1. We're reviewing a transaction
2. There's an error

We shouldn't let the user submit a transaction if there are errors.

src/components/InputV2/Button.tsx
```diff
@@ -0,0 +1,72 @@
...
+  gray: PropTypes.bool,
+  red: PropTypes.bool,
+  green: PropTypes.bool
+}
+
+const getButtonClassName = (props: ButtonClassNameProps): string => {
+  return Object.keys(props)
+    .filter(name => props[name] && buttonClassNames.includes(name))
+    .join(' ')
+}
```
navFooh, line 32: @Schwartz10 just FYI,

This part converts the passed props to HTML classnames, which will allow us to refactor `ButtonV2` completely so it uses the same styling from `base-css`. Then it doesn't need to be a styled component anymore.

However, we first need to phase out all the old `Button` components and remove any styled system declarations from the `ButtonV2` components. Then we can apply the button styling to `button` elements and `a.button` elements (for `ButtonV2Link`). Currently this is limited to `input[type="button"]` so we don't mess with existing components.

src/components/Link/SmartLink.tsx
```diff
@@ -31,21 +31,26 @@ export function SmartLink({
   )
 }
 
-interface SmartLinkProps {
-  href: string
+export interface SmartLinkProps {
   children: ReactNode
+  href?: string
```
navFooh, line 36: @Schwartz10 I made `href` optional, as it's not required when using the `download` attribute

src/customPropTypes.ts
```diff
@@ -74,18 +98,19 @@ export const FILECOIN_NUMBER_PROPTYPE: Requireable<any> = Object.assign(
...
-  timestamp: oneOfType([string, number]).isRequired,
-  method: string.isRequired,
-  params: object.isRequired
+  nonce: number.isRequired,
+  method: number.isRequired,
+  value: BIGNUMBER_PROPTYPE.isRequired,
+  gasPremium: BIGNUMBER_PROPTYPE.isRequired,
+  gasFeeCap: BIGNUMBER_PROPTYPE.isRequired,
+  gasLimit: number.isRequired,
+  params: oneOfType([string, arrayOf(string)])
```
navFooh, line 113: `MESSAGE_PROPS` didn't seem to be in use so I repurposed it for a PropType to match the `Message` object

src/customPropTypes.ts
```diff
@@ -143,14 +168,18 @@ export const MSIG_METHOD_PROPTYPE = oneOf(
...
-  AwaitingConfirmation = 'AwaitingConfirmation',
-  MPoolPushing = 'MPoolPushing'
+  LoadingMessage = 0,
+  FillingForm,
+  FillingTxFee,
+  LoadingTxFee,
+  LoadingTxDetails,
+  AwaitingConfirmation,
+  MPoolPushing
 }
```
navFooh, line 183: @Schwartz10 see comment

src/components/InputV2/Base.tsx
navFooh: sounds good to me!

src/components/Transaction/Form.tsx
navFooh: `txFee` is a passed as a prop so that the parent component (`Send.tsx`) can access the value.
I'm not sure how we'd achieve that with `useMemo` in this case

src/components/Transaction/Form.tsx
navFooh: good to know!

src/components/Transaction/Form.tsx
navFooh: that makes a lot of sense!

src/components/Transaction/Form.tsx
navFooh: Hmm yeah I implemented this for the safe.

So for getting the gas parameters, we first create the base64 params with a nonce of `0`, as the nonce is encoded in the parameters. When we actually send the message, we get the nonce and need to recreate the base64 params with the actual nonce. This component can't be aware of how the params are created because it's different for each msig method.

The only way we can remove this prop I think is to get the nonce in the parent component already and pass it to the Form component. The caveat is that the nonce might be invalid if multiple tx windows were opened at the same time.

src/components/Transaction/Form.tsx
navFooh: There are two states for the next button.

In the first state, when we are filling the form, `message` will be null when a form value is invalid. So then the Next button (labeled Review) will be disabled.

In the next state, we are filling / calculating the tx fee. When there is no valid tx fee, `messageWithGas` will be null and the Next button (labeled send) will be disabled.

At least, that's the logic behind here, let me know if it doesn't work as expected.

src/components/Transaction/Form.tsx
navFooh: And FYI, the reason that `Send.tsx` needs to access the wallet is because it determines the Total value, which is specific per implementation. In the wallet, total value for Withdraw does _not_ include the tx fee

PR #351, opened 2022-05-27
Tx fee fixes

navFooh:


navFooh:
@Schwartz10 small update to hide all the tx fee interface and total amount when going back to filling the tx form

PR #362, opened 2022-05-28
Update/connect screens

Schwartz10:
@navFooh would appreciate a review here.

The UI isn't _perfect_ like you would make it, but it at least aligns closer with the other designs. The create seed phrase option is still the furthest away from the most congruent UI, but this is also the least popular wallet selection option (i thought about even deleting it because it's dangerous to use).

A couple weird things that I can't seem to piece together:

1. In the Create seed phrase wallet option, the `enter` key doesnt seem to fire the onSubmit handler. I'm guessing this has something to do with there (sometimes) not being an input rendered on the dom. The enter key _does_ work when there _are_ input fields rendered. If something obvious sticks out to you, let me know. If not, we can backlog a ticket to fix this.
2. In the import seed phrase wallet option, the loading screen seems to appear very late. It looks perfectly fine for import private key option, and both are pretty much exactly the same?
3. Having to do with point 2 above - I noticed that the `create-wallet-provider` action gets dispatched twice. I've searched long and hard for the culprit of the _second_ (and unnecessary) dispatch, but I can't find it anywhere. I think https://github.com/facebook/react/issues/16295 is the most obvious explanation, and our reducer is pure, so this shouldnt actually cause any noticeable differences in our app, and shouldn't happen at all in production. 

One other observation that I'm punting on for now - when importing private keys, the `create-wallet-provider` action gets dispatched twice. Not sure why that's happening. 

It's probably easiest to test this in the wallet

[EDIT]: correct "the UI is perfect.." => "the UI isn't perfect" :D


src/components/Connect/Burner/CreateSeed/index.tsx
```diff
@@ -27,14 +26,13 @@ const Create: FC<{
...
   const [loading, setLoading] = useState(true)
-  const [returningHome, setReturningHome] = useState(false)
   const [canContinue, setCanContinue] = useState(false)
   const [importSeedError, setImportSeedError] = useState(false)
-  const [acceptedWarning, setAcceptedWarning] = useState(false)
   const { dispatch, fetchDefaultWallet, walletList, lotusApiAddr } =
     useWalletProvider()
 
   const nextStep = () => {
+    console.log('hi')
```
navFooh: don't forget to remove this one :)

src/components/Connect/Burner/CreateSeed/index.tsx
```diff
@@ -78,81 +76,66 @@ const Create: FC<{
...
-            <>
-              {loading || walkthroughStep === 4 ? (
-                <LoadingScreen />
+      {loading || walkthroughStep === 4 ? (
+        <LoadingScreen />
+      ) : (
+        <form
+          onSubmit={e => {
+            e.preventDefault()
+            console.log('in here')
```
navFooh: and this one

src/components/InputV2/Base.tsx
```diff
@@ -76,7 +79,8 @@ export const BaseInput = ({
   </Label>
 )
 
-export interface BaseInputProps {
+export interface BaseInputProps
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
```
Schwartz10: Here i omit the `onChange` type because we override it with our own version of `onChange`

PR #368, opened 2022-05-31
Update for Connect Screens PR

navFooh:


src/components/Connect/Burner/CreateSeed/Walkthrough/RevealMnemonic.tsx
```diff
@@ -112,14 +111,9 @@ const Reveal = ({ mnemonic, valid }) => {
...
-          return (
-            /* eslint-disable react/no-array-index-key */
-            <MenuItem key={i}>
-              <Word num={i + 1} word={word} valid={valid} />
-            </MenuItem>
-          )
-        })}
+        {mnemonic.split(' ').map((word, i) => (
+          <Word key={i} num={i + 1} word={word} valid={valid} />
+        ))}
```
navFooh, line 116: Wrapping it in the extra `MenuItem` only complicated the flex layout behaviour

src/components/Connect/Burner/CreateSeed/Walkthrough/WordPrompt.tsx
```diff
@@ -49,29 +48,19 @@ const WordPrompt = ({
...
         )}
       </Box>
-      <MnemonicWordContainer
-        display='flex'
-        alignItems='center'
-        justifyItems='center'
-        flexWrap='wrap'
-        mt={3}
-      >
+      <MnemonicWordContainer>
```
navFooh, line 51: `MnemonicWordContainer` itself is styled so this isn't necessary

src/components/Connect/Burner/CreateSeed/index.stories.jsx
```diff
@@ -15,12 +15,10 @@ export default {
...
-    </>
-  )
-}
+const Template = args => <CreateSeed {...args} />
 
 export const Base = Template.bind({})
+Base.args = {
+  back: () => console.log('Clicked Back'),
+  next: () => console.log('Clicked Next')
+}
```
navFooh, line 24: Missing props here that should be required. It didn't give an error before because the component's proptypes weren't set

src/components/Connect/Burner/CreateSeed/index.test.js
```diff
@@ -2,7 +2,7 @@ import { cleanup, render, screen, act, fireEvent } from '@testing-library/react'
 import composeMockAppTree from '../../../../test-utils/composeMockAppTree'
 import { flushPromises } from '../../../../test-utils'
 
-import Create from '.'
+import { CreateSeed } from '.'
```
navFooh, line 5: I don't like the default exports because it leaves room for the implementation to import it as whatever, creating discrepancies in component names

src/components/Connect/Burner/CreateSeed/index.tsx
```diff
@@ -73,78 +70,62 @@ const Create: FC<{
...
-      )}
-    </>
+            Back
+          </ButtonV2>
+          <ButtonV2 large green type='submit'>
+            {walkthroughStep === 1 ? "I've recorded my seed phrase" : 'Next'}
+          </ButtonV2>
+        </ButtonRowSpaced>
+      </form>
+    </WideDialog>
```
navFooh, line 112: I think the main takeaways here are that `OnBoardCard` could be deprecated in favor of `ShadowBox`. Also the custom styled `Box` wrapper wasn't necessary here. I added a `WideDialog` component that is `50em` instead of `35em` which can be used here.

src/components/Connect/Burner/CreateSeed/index.tsx
```diff
@@ -73,78 +70,62 @@ const Create: FC<{
...
+  back: () => void
+  next: () => void
 }
 
-Create.defaultProps = {
-  initialWalkthroughStep: 1
+CreateSeed.propTypes = {
+  initialWalkthroughStep: PropTypes.number,
+  back: PropTypes.func.isRequired,
+  next: PropTypes.func.isRequired
```
navFooh, line 126: These were missing

src/components/Connect/Burner/ImportPk/index.stories.jsx
```diff
@@ -0,0 +1,24 @@
...
+  parameters: { actions: { argTypesRegex: '^on.*' } }
+}
+
+const Template = args => <ImportPk {...args} />
+
+export const Base = Template.bind({})
+Base.args = {
+  back: () => console.log('Clicked Back'),
+  next: () => console.log('Clicked Next')
+}
```
navFooh, line 24: `ImportPk` story was missing

src/components/Connect/Burner/ImportPk/index.tsx
```diff
@@ -8,13 +9,7 @@ import { Dialog, ShadowBox, ButtonRowSpaced } from '../../../Layout'
...
 import { Loading } from '../../Loading'
 
-export default function ImportSeed({
-  back,
-  next
-}: {
-  back: () => void
-  next: () => void
-}) {
+export const ImportPk = ({ back, next }: ImportPkProps) => {
```
navFooh, line 12: Note that the default export here allowed a copy-paste error calling this component `ImportSeed` instead of `ImportPk`

src/components/Connect/Burner/ImportPk/index.tsx
```diff
@@ -63,24 +58,33 @@ export default function ImportSeed({
...
+        </ShadowBox>
+        <ButtonRowSpaced>
+          <ButtonV2 large disabled={loading} onClick={back} type='button'>
+            Back
+          </ButtonV2>
+          <ButtonV2 large disabled={loading} green type='submit'>
+            Connect
+          </ButtonV2>
+        </ButtonRowSpaced>
+      </form>
```
navFooh, line 77: I think it's nicer to have the buttons below the `ShadowBox` like in other similar screens. We could opt for buttons inside the box, but then I think we should do that in other places as well

src/components/Connect/Burner/ImportSeed/index.stories.jsx
```diff
@@ -1,10 +1,10 @@
 import theme from '../../../theme'
 import ThemeProvider from '../../../ThemeProvider'
-import InputMnemonic from './'
+import { ImportSeed } from '.'
```
navFooh, line 3: Default export naming here as well. I guess a seed is the same as Mnemonic, but still bad consistency

src/components/Connect/Ledger/LedgerError.stories.jsx
```diff
@@ -26,12 +26,14 @@ const ledgerState = newLedgerState => ({
 })
 
 export default {
-  title: 'LedgerError/LedgerError',
+  title: 'Connect/LedgerError',
```
navFooh, line 29: I moved all Connect type things to the Connect story group

src/components/Connect/Ledger/LedgerError.stories.jsx
```diff
@@ -26,12 +26,14 @@ const ledgerState = newLedgerState => ({
...
   component: LedgerError,
   decorators: [
     Story => (
       <div style={{ display: 'flex', justifyContent: 'center' }}>
-        <ThemeProvider theme={theme}>{Story()}</ThemeProvider>
+        <ThemeProvider theme={theme}>
+          <Dialog>{Story()}</Dialog>
```
navFooh, line 35: Wrapped the `LedgerError` in a `Dialog` so it lays out like it would in the `ConnectLedger` component

src/components/Connect/Ledger/LedgerError.tsx
```diff
@@ -1,46 +1,29 @@
...
-`
+export const LedgerError = ({ ...errors }: LedgerErrorProps) =>
+  hasLedgerError({ ...errors }) && (
+    <ErrorBox>
+      <p>{reportLedgerConfigError({ ...errors })}</p>
+      {errors.inUseByAnotherApp && (
+        <p>(Most of the time, this is Ledger Live!)</p>
+      )}
+    </ErrorBox>
+  )
```
navFooh, line 18: LedgerError can just be a normal `ErrorBox`, the parent should wrap it in a `Dialog`

src/components/Connect/Ledger/LedgerError.tsx
```diff
@@ -1,46 +1,29 @@
...
 
-const Helper: FC<HelperProps> = ({ ...errors }) => (
-  <Box>
```
navFooh, line 21: Also no need for the `Box` wrapper then

src/components/Connect/Ledger/index.stories.jsx
```diff
@@ -16,7 +15,10 @@ export default {
...
 
-const Template = args => <Ledger {...args} />
+const Template = args => <ConnectLedger {...args} />
 
 export const Base = Template.bind({})
-Base.args = {}
+Base.args = {
+  back: () => console.log('Clicked Back'),
+  next: () => console.log('Clicked Next')
+}
```
navFooh, line 24: Also here missing props that showed when adding proptypes

src/components/Connect/Ledger/index.test.js
```diff
@@ -3,17 +3,19 @@ import { initialState as wpInitialState } from '../../../services/WalletProvider
...
+import { ConnectLedger } from '.'
 import { flushPromises } from '../../../test-utils'
 import { TESTNET_PATH_CODE } from '../../../constants'
 import createPath from '../../../utils/createPath'
 
 describe('Ledger configuration', () => {
   let backSpy, nextSpy
-  afterEach(() => {
-    jest.clearAllMocks()
+  beforeEach(() => {
```
navFooh, line 13: `backSpy` and `nextSpy` were initially passed as `undefined` because they were set in `afterEach`

src/components/InputV2/Base.tsx
```diff
@@ -79,8 +78,7 @@ export const BaseInput = ({
   </Label>
 )
 
-export interface BaseInputProps
-  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
```
navFooh, line 83: I wouldn't add all of the HTML input attributes.

Apart from onChange, we were already specifying other props that overlapped with the attributes (`disabled`, `placeholder`, `autoFocus`, etc.)

But mostly, the root element of this component is not an HTML input element and there would be quite a few attributes that typescript would then allow to pass that we are not using / passing to our input element at all.

Since we're just adding the `autoComplete`, I think it'd be best to do just add that now. We can change it to accept all intput attributes but then we'd have to use rest props and object spread them into the `input`

src/components/InputV2/Base.tsx
```diff
@@ -90,7 +88,8 @@ export interface BaseInputProps
   error?: string
   name?: string
   type?: string
-  autofocus?: boolean
+  autoFocus?: boolean
```
navFooh, line 91: My bad, but I think good to fix right now

src/components/InputV2/Button.tsx
```diff
@@ -44,7 +45,7 @@ export const ButtonInput = ({
   ...baseInputProps
 }: ButtonInputProps) => (
   <BaseInput
-    type='button'
+    type={submit ? 'submit' : 'button'}
```
navFooh, line 48: Added a submit type input button, which makes sense to have

src/components/InputV2/PrivateKey.tsx
```diff
@@ -15,9 +15,11 @@ export const PrivateKeyInput = ({
   ...baseProps
 }: PrivateKeyInputProps) => {
   const [hasFocus, setHasFocus] = useState<boolean>(false)
+  const [hasChanged, setHasChanged] = useState<boolean>(false)
```
navFooh, line 18: Omitting this means the input would have a different behaviour around error handling than the other ones. We added this so that we only show errors after the user actually used the input, otherwise it would seem like inputs had errors way too fast.

Same for `SeedPhraseInput`

src/components/InputV2/PrivateKey.tsx
```diff
@@ -15,9 +15,11 @@ export const PrivateKeyInput = ({
...
 
   // Check for input errors
   const error = useMemo<string>(() => {
+    if (importError) return 'Error importing private key'
```
navFooh, line 22: Moved this up, because I think we should prioritize this error when the parent passes it.

src/components/InputV2/PrivateKey.tsx
```diff
@@ -71,20 +72,24 @@ export const PrivateKeyInput = ({
...
 
-const { error, type, placeholder, ...addressProps } = BaseInputPropTypes
+const { error, type, autoComplete, placeholder, ...privateKeyProps } =
+  BaseInputPropTypes
 
 PrivateKeyInput.propTypes = {
   setIsValid: PropTypes.func,
-  ...addressProps
+  importError: PropTypes.string,
+  ...privateKeyProps
```
navFooh, line 92: Added missing proptype, same for `SeedPhraseInput`

src/components/InputV2/PrivateKey.tsx
```diff
@@ -71,20 +72,24 @@ export const PrivateKeyInput = ({
...
 
 export type PrivateKeyInputProps = {
   setIsValid?: (isValid: boolean) => void
   importError?: string
-} & Omit<BaseInputProps, 'error' | 'type' | 'placeholder'>
+} & Omit<BaseInputProps, 'error' | 'type' | 'autoComplete' | 'placeholder'>
 
-const { error, type, placeholder, ...addressProps } = BaseInputPropTypes
+const { error, type, autoComplete, placeholder, ...privateKeyProps } =
+  BaseInputPropTypes
```
navFooh, line 87: autoComplete is also set locally, so we need to omit it from the BaseInputProps. Same for `SeedPhraseInput`

src/components/InputV2/SeedPhrase.stories.jsx
```diff
@@ -0,0 +1,45 @@
...
+  ],
+  parameters: { actions: { argTypesRegex: '^on.*' } }
+}
+
+const Template = args => <StoryComponent {...args} />
+
+export const Base = Template.bind({})
+Base.args = {
+  label: 'Enter a seed phrase'
+}
```
navFooh, line 45: Added missing story

src/components/Layout/Boxes.jsx
```diff
@@ -109,8 +109,8 @@ const ErrorBoxEl = styled(BoxBase)`
   }
 `
 
-export const ErrorBox = ({ children, ...props }) => (
-  <ErrorBoxEl {...props}>
+export const ErrorBox = ({ children }) => (
+  <ErrorBoxEl>
```
navFooh, line 113: Rather wouldn't do this when there is no (good) reason

src/components/Layout/Boxes.jsx
```diff
@@ -5,7 +5,7 @@ import { IconWarn, IconError } from '../Icons'
 /**
  * Box base styling
  */
-export const BoxBase = styled.div`
+const BoxBase = styled.div`
```
navFooh, line 8: We _could_, but should we create there the possibility for custom boxes in other apps? Afraid it could get messy, currently there is no need

src/components/Layout/Dialogs.jsx
```diff
@@ -14,3 +14,7 @@ export const Dialog = styled.div`
     gap: ${space()};
   }
 `
+
+export const WideDialog = styled(Dialog)`
+  max-width: 50em;
+`
```
navFooh, line 20: Nice for the seed phrase creation, maybe other dialogs later as well

src/components/MnemonicWord/Circle.ts
```diff
@@ -1,15 +1,9 @@
...
-}))`
-  display: flex;
-  justify-content: center;
-  align-items: center;
+export default styled.div`
+  width: 1.5em;
+  height: 1.5em;
+  font-size: 1.125em;
+  text-align: center;
+  line-height: 1.5em;
```
navFooh, line 8: I got some weird DOM behaviour around the Mnemonic word part, so thought it could use a quick clean up.

PR #371, opened 2022-06-01
Add Select input

navFooh:


src/components/InputV2/Select.tsx
```diff
@@ -0,0 +1,132 @@
...
+            </option>
+          ))}
+        </select>
+        <div className='select-arrow' />
+      </div>
+    </Label>
+  )
+}
+
+export interface SelectProps {
```
Schwartz10, line 77: Should we not just extend the normal Select props through React here rather than redefine the props on our own?

We could override the `onChange` handler type?

src/components/Transaction/Header.tsx
```diff
@@ -44,6 +45,9 @@ export const TransactionHeader = ({
         txState === TxState.LoadingTxDetails ||
         txState === TxState.MPoolPushing) && <LoaderGlyph />}
     </StandardBox>
+    {warning && txState < TxState.LoadingTxDetails && (
+      <WarningBox>{warning}</WarningBox>
+    )}
```
Schwartz10: Thinking through the UX on this... are we sure this is the right state to display the warning? I'd probably vote for one of the following:

1. Always show the warning if there is one (state agnostic)
2. Show the warning during the transaction review state

Schwartz10 (commented):
Hey @navFooh see my couple comments. The select component looks good to me!

I'm wondering... why do all the other input fields' snapshots change in this PR? Looking at the diff ([example](https://github.com/glifio/react-components/pull/371/files#diff-11e1cb2c26c2ea0a2182d46f60414213afef7ca944c9aeaa25a9374683d13eb6R202)), it looks like the select components classes are generally added to a stylesheet, but those specific classnames are not applied anywhere. does that sound right to you?

src/components/Transaction/Header.tsx
navFooh: Good call! I think 2 would be good UX, let's try it out and see how it feels in the safe.

src/components/InputV2/Select.tsx
navFooh: Concerning the snapshots, I'm not 100% sure how the renderer works, but it doesn't seem to care about what is in the HTML for what CSS it will render. The CSS for all buttons seems to be the same, regardless of whether they contain a unit, delete button, or `select-wrapper`  in this case. I think it only renders the CSS here due to the use of styled components. Note that styles coming from `base-css` aren't rendered here either.

Do you mean extending the `BaseInputProps`?

I realize there is a bit of duplication in the props and how they are used, but extending the `BaseInputProps` for `SelectProps` feels wrong to me. Because the `Select` component does not inherit from `BaseInput`, they shouldn't necessarily affect eachother. Say that we use the `BaseInputProps`, strip what we don't need for `Select` and add what we do need, that might save a few lines of code now. But if we modify `BaseInput` later, that means we always have to be aware of how that affects `Select`, which feels bad for unrelated components. It's different for components like `NumberInput` which actually extend `BaseInput`.

navFooh:
Will merge and publish so we can try it out in Safe!

PR #380, opened 2022-06-06
Select number

navFooh:
Added a select input for any number and a range of numbers


Schwartz10 (approved):
I didn't drill into the logic _too_ much, but the UI looks great and i can't see any issues with logic. Will be easier for me to find when I test in tandem with safe and wallet

PR #393, opened 2022-06-13
Remove deprecated components, update Tx.Form

navFooh:


Schwartz10:
If you rebase this off https://github.com/glifio/react-components/pull/394 after merging, it should pass

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,25 +1,17 @@
...
+import { DetailCaption } from '../detail'
 import convertAddrToPrefix from '../../../utils/convertAddrToPrefix'
 
-const ViewState = styled(P).attrs(() => ({
-  color: 'core.primary',
-  role: 'button'
-}))`
-  cursor: pointer;
-`
-
```
navFooh, line 22: @Schwartz10 `p[role="button"]` styling has been added to `base-css` because it occurs in a few places. This makes these custom styled components redundant

src/components/HistoryTables/MessageHistory/Detail.tsx
```diff
@@ -29,16 +26,13 @@ import {
...
 // add RelativeTime plugin to Day.js
 dayjs.extend(relativeTime.default)
 
-const SeeMore = styled(P).attrs(() => ({
-  color: 'core.primary',
-  role: 'button'
-}))`
-  cursor: pointer;
+const SpanGray = styled.span`
+  color: var(--gray-medium);
```
navFooh, line 35: Was wondering is we should make this something generic, but it only lives here for now

src/components/HistoryTables/MessageHistory/Detail.tsx
```diff
@@ -95,14 +89,20 @@ export default function MessageDetail(props: MessageDetailProps) {
...
+          be shown here in 1-2 minutes.
+        </InfoBox>
+      ) : !loading && !error && !message ? (
+        <StandardBox>
+          <p>Message {cid} not found.</p>
+          <p>
+            Note - it may take 1-2 minutes for a recently confirmed message to
+            show up here.
+          </p>
+        </StandardBox>
```
navFooh, line 103: Deprecated Cards in favor of the new boxes

src/components/HistoryTables/detail.tsx
```diff
@@ -180,100 +179,6 @@ DetailCaption.propTypes = {
...
-
-type AddressLineProps = {
-  value: string | Address
-} & LineProps
-
-AddressLine.propTypes = {
-  value: PropTypes.oneOfType([PropTypes.string, ADDRESS_PROPTYPE]).isRequired,
-  ...LinePropTypes
-}
-
```
navFooh, line 276: @Schwartz10 The `LineWrapper` and `Line` seem to be useful in other contexts and are Layout related, so I moved them to `Layout`

src/components/InputV2/Base.tsx
```diff
@@ -1,6 +1,61 @@
...
+        `}
+
+      ${props =>
+        props.disabled &&
+        css`
+          color: var(--gray-dark) !important;
+        `}
+    }
+  }
+`
```
navFooh, line 58: @Schwartz10 For all of the inputs that have specific styling requirements in `Label` (like `.button-wrapper`, `.unit-wrapper`), I created their own specific `Label` extension. So the styling definition is now closer to the HTML where they apply, which is nice. But also the styles won't be applied to elements that don't need it. You can see by the classes that don't apply to specific elements being removed from the snapshots

src/components/InputV2/Label.tsx
```diff
@@ -45,86 +45,4 @@ export const Label = styled.label`
...
-        props.disabled &&
-        css`
-          border-color: var(--gray-dark);
-        `}
-    }
-  }
-
-  .info-wrapper {
-    padding: 0.7em 1em;
-  }
```
navFooh, line 129: Meaning all these don't need to be defined here. Now it only contains what applies to all Label instances

src/components/Transaction/Fee.tsx
```diff
@@ -46,6 +46,7 @@ export const TransactionFee = ({
   return (
     txState >= TxState.FillingTxFee && (
       <>
+        <hr />
```
navFooh, line 49: This extra line looks nice in the Tx Review so that it separates the UI that was just added from the one that was already there

src/components/Transaction/Total.tsx
```diff
@@ -15,13 +15,13 @@ const Total = styled.div`
 
 export const TransactionTotal = ({ total }: TransactionTotalProps) => {
   return (
-    <footer>
+    <>
```
navFooh, line 18: Removing the footer container makes the style more consistent now it's contained in the `Lines` wrapper

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -99,13 +91,13 @@ export function ActorState({ address }: { address: string }) {
             <Box minWidth='200px' flex='0 1 25%'>
               State
             </Box>
-            <ViewState onClick={() => setViewActorState(!viewActorState)}>
+            <p role='button' onClick={() => setViewActorState(!viewActorState)}>
               Click to{' '}
               {viewActorState ? 'hide actor state ↑' : 'see actor state ↓'}
-            </ViewState>
+            </p>
```
Schwartz10: I think this might break google translator apis - the conditional rendering should conditionally render two separate `p` tags

src/components/HistoryTables/MessageHistory/Detail.tsx
Schwartz10: Think it's fine for now!

src/components/HistoryTables/MessageHistory/Detail.tsx
Schwartz10: LFG

src/components/HistoryTables/MessageHistory/Detail.tsx
```diff
@@ -165,15 +165,15 @@ export default function MessageDetail(props: MessageDetailProps) {
                 </Line>
               )}
               <hr />
-              <SeeMore onClick={() => setSeeMore(!seeMore)}>
+              <p role='button' onClick={() => setSeeMore(!seeMore)}>
                 Click to see {seeMore ? 'less ↑' : 'more ↓'}
-              </SeeMore>
+              </p>
```
Schwartz10: Same thing here with google translate API - i might be wrong, but better safe than sorry?

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -195,9 +187,9 @@ export default function ProposalDetail(props: ProposalDetailProps) {
           depth={0}
         />
         <hr />
-        <SeeMore onClick={() => setSeeMore(!seeMore)}>
+        <p role='button' onClick={() => setSeeMore(!seeMore)}>
           Click to see {seeMore ? 'less ↑' : 'more ↓'}
-        </SeeMore>
+        </p>
         <hr />
```
Schwartz10: same conditional note

src/components/InputV2/Base.tsx
Schwartz10: Good thinking!

src/components/Transaction/Buttons.tsx
```diff
@@ -8,17 +8,25 @@ export const TransactionButtons = ({
   nextDisabled,
   backText,
   nextText,
+  backType,
+  nextType,
```
Schwartz10: curious - why do these need to be passed as props instead of hardcoded in this context? is the next button not always `submit` and the back button not `button`

Judging from you adding `onClickNext` - i'm sure i'm not thinking of some instance of needing this in one of our apps

src/components/Transaction/Total.tsx
Schwartz10: So nice!

src/components/index.ts
```diff
@@ -38,7 +35,7 @@ export { default as ProposalDetail } from './HistoryTables/MsigProposals/Proposa
 export * from './HistoryTables/ActorState'
 export * from './HistoryTables/PendingMsgContext'
 export * from './HistoryTables/methodName'
-export { LineWrapper, Line, Parameters } from './HistoryTables/detail'
```
Schwartz10, line 41: Feels like we might as well export the new lines component somewhere in here since they'll presumably be useful to us down the road? doesn't add any weight to the package anyway?

Schwartz10 (commented):
Hey @navFooh so nice deleting all this code! A couple small comments

Note because i'm traveling this week, i haven't run the new transaction form element changes linked to the wallet or safe, so i haven't seen the new UIs yet. I trust your judgement though, and i'm down to just tweak down the road if we see something after merging.

See my few tiny comments, and then feel free to squash and merge tomorrow once you've taken a look

src/components/HistoryTables/ActorState/index.tsx
navFooh: You are totally right, I just checked by translating to Dutch and this breaks the Google Translation indeed (it doesn't update when toggling the button). Fixed it by rendering the two separate `p`s

src/components/HistoryTables/MessageHistory/Detail.tsx
navFooh: You were not :) good to know for future implementations!

src/components/Transaction/Buttons.tsx
navFooh: Yeah I think you are right, we can make this less versatile and more tailored to our use case, so it encourages using the buttons in a form with the submit action

src/components/index.ts
navFooh: It is exported in `src\components\Layout\index.js`

PR #394, opened 2022-06-13
Update jest deps

Schwartz10:
multiformats module failing test


Schwartz10:
@navFooh this fixes the `npm i` problem but it introduces a new one, which I'm still investigating the best way to fix.

Basically, jest tries to import the esm export of the `multiformats/cid` module, which it chokes on. If i manually import the cjs export of the multiformats module directly, the tests pass. For example, from `validateCID.ts`:

```
import { CID } from '../../node_modules/multiformats/cjs/src/cid'
```

It seems like we're _not_ using babel-jest to transform `/node_modules/` [here](https://github.com/glifio/react-components/blob/primary/jest.config.js#L13). I've tried a couple ways of ignoring all node_modules _except_ `multiformats/cid` but I haven't seemed to get it to like my configuration (i'm almost positive i'm doing something wrong). Not ignoring the node_modules just takes forever to run the tests, so that's not a great option.

Another option I thought would be to use the [`moduleNameMapper`](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring) to manually point `multiformats/cid` directly to the cjs implementation, but the multiformats/cjs doesn't have its own package.json, but since the cjs directory doesn't have its own package.json file, the moduleNameMapper hasn't been able to "find it".

Consider this my notes so far, i'm going to keep experimenting a bit until something works

Schwartz10:
Another potential solution path https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs

Schwartz10:
(just leaving potential solutions here as i find them) https://github.com/facebook/jest/issues/9771#issuecomment-890454033

Schwartz10:
Ok @navFooh I have _a_ fix, albeit it does not feel like the best one bc it's hacky and requires 1 by 1 fixes for any module that exports esm modules. The idea is to basically manually require the cjs module for node modules that default to esm, since esm support is experimental in jest at the moment... 

However, I'm confused at what caused this to happen in the first place - (why would updating jest remove support for esm?). This makes me feel like i dont have a deep enough understanding of the problem in the first place.

Anyways, if you feel ok about this solution so do i. feel free to merge if you want.

This PR also removes an old component that we no longer needed (i'm guessing your cleanup PR did that too)

navFooh:
I haven't been able to find a better solution so far. For now I think we should go with it.

PR #407, opened 2022-06-16
Decode actor cid

navFooh:
Noticed the following TypeScript error in some of the tests:
image

Changing the `WalletProviderWrapper` children property type from `ReactChildren` to `ReactNode` should fix this


src/services/WalletProvider/index.tsx
```diff
@@ -4,7 +4,7 @@ import {
   useReducer,
   useCallback,
   Dispatch,
-  ReactChildren,
+  ReactNode,
```
navFooh, line 7: @Schwartz10 See PR description

src/utils/decodeActorCID.ts
```diff
@@ -1,14 +1,14 @@
...
     const decoded = base32Decode(cid.slice(1).toUpperCase(), 'RFC4648')
     const actorCode = new TextDecoder('utf-8').decode(decoded.slice(4))
     if (!actorCode.includes('fil/')) {
-      return 'Unknown actor'
+      throw new Error('unknown actor code')
     }
     return actorCode
   } catch {
-    return 'Unknown actor'
+    throw new Error('invalid actor code')
```
navFooh, line 12: @Schwartz10 This is basically the way the util was implemented in the Safe.

It seemed to make sense to use the `react-components` implementation in the Safe, but probably better to throw an error so that invalid actors can be caught instead of just displaying `Unknown actor`. Not sure how the implementation of custom actors will affect all this, but I guess we will cross that bridge when we get there

package.json
```diff
@@ -87,19 +86,14 @@
...
     "lodash.clonedeep": "^4.5.0",
     "multiformats": "^9.6.5",
-    "path": "^0.12.7",
     "prop-types": "^15.7.2",
     "styled-system": "^5.1.5",
-    "subscriptions-transport-ws": "^0.11.0",
-    "swr": "^1.0.1",
-    "trim": "^1.0.1",
-    "ws": "^8.4.2"
+    "swr": "^1.0.1"
```
navFooh, line 96: Also tidies up the dependencies

Schwartz10 (approved):

PR #416, opened 2022-06-21
Smartlink retain params

navFooh:
This makes sure that SmartLink retains some default query params (`address`, `msigAddress`) and gives the option to retain others as well. I think we need some form of this for sure to skip the safe selection screen, but let's discuss in a call


src/components/Link/SmartLink.tsx
```diff
@@ -1,33 +1,67 @@
...
+  const retainParams = [
+    ...(retainDefaultParams ? defaultParams : []),
+    ...(retainExtraParams ? retainExtraParams : [])
+  ]
+  const hrefWithParams = useMemo<string>(() => {
+    // Don't alter absolute URLs or when not retaining params
+    if (isAbsolute || !retainParams.length) return href
+
+    // Get query params from passed href
+    const [path, paramString] = href.split('?')
```
Schwartz10: I think we could use a built-in class for this to be safe like `URL`? Probably fine.

Schwartz10 (approved):

PR #426, opened 2022-06-23
Set up fix for https://github.com/glifio/safe/issues/104

Schwartz10:


navFooh:
LGTM! Just wondering, as an alternative, would it be a solid solution to check something like this

```
if (newMessage.to !== EXEC_ACTOR)
  pushPendingMessage(
    newMessage.toPendingMessage(msgCid['/']) as MessagePending
  )
```

We'd have to add `EXEC_ACTOR` to the `constants.ts` file in `react-components`. But I understand if we might want to app to have control over this or if this check could not be reliable in some cases

Schwartz10:
@navFooh good thought! The only issue is that we don't _always_ want to stop actor init transactions from being pushed to the pending message cache - there are specific cases where doing so screws up the message history (creating a safe from the safe app) but in the future this could change (for example, creating a new actor from the wallet we _would_ push the pending message because _that_ tx does show up in my wallet's tx history)

PR #434, opened 2022-06-27
Custom scalar types

navFooh:
The custom scalars in the GraphQL schema, `Int64` and `Uint64`, were treated as `any` type in TypeScript. This caused us to wrongly interpret them as string, while they are actually sent as numbers.

I'm not sure if `Int64` and `Uint64` _should_ be sent as numbers, as the range for a `number` type in Javascript is probably less than a `Int64` or `Uint64` in Go. However, I feel that the values that could get out of range are explicitly sent as `string` anyway.


navFooh:
This also made me notice that there are 5 different data types defined for height in the GraphQL schema. Not sure if they should be different based on context, but seems a bit inconsistent:

```
Actor.Height: String
Block.Height: Int64
ChainHead.height: Int64
Message.height: Float
MessageConfirmed.height: Int64
MessagePending.height: String
QueryBlockArgs.height: Int64
QueryMessageArgs.height: Int
TipSet.height: Uint64
```

Especially between `MessageConfirmed` and `MessagePending` this is unexpected, as the custom `useMessage` hook can return either. Then `height` is a different data type depending on whether it's one or the other.

src/components/HistoryTables/MessageHistory/useAge.ts
```diff
@@ -20,7 +20,7 @@ export const useUnformattedDateTime = (
 
   useChainHeadSubscription({
     variables: {},
-    skip: !message,
+    skip: !message || !!message?.block?.Timestamp,
```
Schwartz10, line 24: Just want to make sure this is intentional and doesn't introduce unexpected behavior - 

this says to me "dont subscribe to chain head if the message has a timestamp"

Do we know for sure that `!!message?.block?.Timestamp` evaluates to truthy/falsey when we expect it to? One way to test this is just submit a transaction, open it up immediately, and watch it go from pending -> confirmed state

then open a new table on the explorer, and refresh _immediately_ after the message confirms in original UI step above ^^. Both should show a correct looking time (they won't be exactly the same, but close enough is fine)

src/components/HistoryTables/types.ts
```diff
@@ -15,7 +15,7 @@ export type MessageConfirmedRow = Pick<
 
 export type MessagePendingRow = Pick<
   MessagePending,
-  'cid' | 'method' | 'from' | 'to' | 'value'
+  'cid' | 'method' | 'height' | 'from' | 'to' | 'value'
```
Schwartz10: This change doesn't really make sense to me - a pending message doesn't have a height because that would imply it's confirmed

Schwartz10 (commented):
Hey @navFooh this mostly looks good but just check my comments - good catch on the typings.

The different types on the backend sounds bad and like something that should get standardized across calls, but I'm not sure if @ganzai-san is already on it with his refactor (it might be the case). So let's hold off on doing anything about it right now. Mabye make a ticket in the graph repo for now so we dont forget?

src/components/HistoryTables/MessageHistory/useAge.ts
navFooh: Well it's a bit funny, when I view the transaction locally it's always a bit more slow than when I have the same transaction open on the wallet at https://wallet-calibration.glif.link/

But both of them actually first show ~1 year ago and then jump to the actual time. On the remote version, this jump goes a lot faster. Undoing my change above doesn't have any influence on this.

The reason for the change is that we do the exact same check in the return statement:
```
if (!!message?.block?.Timestamp) {
  return dayjs.unix(message.block?.Timestamp)
}
```

So if `!!message?.block?.Timestamp` is truthy, we don't use the result of the subscription anyway

src/components/HistoryTables/types.ts
navFooh: That is interesting, according to the schema it does, as a string (see my comment about the height types above).

Anyway, it's fine to remove, but then we have to adjust the prop types passed to `useAge` a bit to make typescript happy, see my latest commit

navFooh: Just did a test and height seems to return an empty string for PendingMessage

navFooh: https://github.com/glifio/graph/issues/67

Schwartz10 (approved):

PR #439, opened 2022-06-29
Refactor

Schwartz10:


.storybook/main.js
```diff
@@ -1,10 +1,30 @@
+const webpack = require('webpack')
+
 module.exports = {
```
Schwartz10, line 3: I mustve merged a PR that merged next > 12, which uses webpack 5. So i had to update storybook to work with webpack5.

Webpack5 does not include polyfills for certain packages that webpack4 shimmed by default. So this webpack config just shims the right things to get storybook to work again

src/utils/isAddrEqual/index.ts
```diff
@@ -2,8 +2,8 @@ import { Address } from '../../generated/graphql'
 
 export const isAddrEqual = (address: Address, addressStr: string): boolean => {
   const noNetworkAddr = addressStr.slice(1)
-  const noNetworkID = address.id?.slice(1)
-  const noNetworkRobust = address.robust?.slice(1)
+  const noNetworkID = address?.id?.slice(1)
+  const noNetworkRobust = address?.robust?.slice(1)
```
Schwartz10, line 6: Just a safe guard in case the server returns undefined since either address is now technically "optional", even though it should always at least return an empty string

src/components/HistoryTables/methodName/index.ts
```diff
@@ -17,7 +17,9 @@ export const getMethodName = (
 ): string => {
   let methodName = ''
 
-  if (actorCode.includes('multisig')) {
+  if (methodNum === 0) {
+    methodName = 'send'
+  } else if (actorCode.includes('multisig')) {
```
Schwartz10, line 22: Fixes small bug where a wallet transfers FIL to a not (yet) existent actor

src/components/AddressLink/index.tsx
```diff
@@ -16,6 +15,17 @@ const Link = styled(SmartLink)`
   }
 `
 
+const AddressWrap = styled.div`
+  display: flex;
+  flexdirection: row;
+  grid-gap: 0.25em;
+  line-height: ${props => props.hideCopy && '1.5'};
+`
```
Schwartz10: @navFooh this seems to have changed some styling in the snapshots - just wanted to flag it in case i missed a class that needs to be manually added when removing the box?

Schwartz10:
image

image

 It looks like some max width properties were removed from things when i moved the box, but the account selector still looks good in the wlalet

.storybook/main.js
```diff
@@ -1,10 +1,25 @@
...
-  "stories": [
-    "../src/**/*.stories.mdx",
-    "../src/**/*.stories.@(js|jsx)"
-  ],
-  "addons": [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials"
-  ]
+  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx)'],
```
navFooh: Should we also add the `ts` and `tsx` extensions here? I think it'd be a good aim to use TypeScript for all our files

package.json
```diff
@@ -66,15 +68,18 @@
     "jest-environment-jsdom": "^28.1.1",
     "jest-styled-components": "7.0.8",
     "mkdirp": "^1.0.4",
-    "next": "^12.0.7",
+    "next": "12.2.0",
```
navFooh: Just wondering, is there a reason to remove the `^` here? As the version will be locked with the package json anyway

src/components/AddressLink/index.tsx
```diff
@@ -26,45 +36,39 @@ export const AddressLink = ({
...
+      <AddressWrap hideCopy={hideCopy}>
         {disableLink ? (
           <span>{linkText}</span>
         ) : (
           <Link color={color} href={linkHref} onClick={onClick}>
             {linkText}
           </Link>
         )}
-        {!hideCopy && address && (
+        {!hideCopy && !!linkText && (
```
navFooh: The reason to check for `address` and not `linkText` is that `linkText` can be a valid string (when we only pass the `id`) while `address` is undefined. Then it will show the copy icon, but the value that will be copied is `"undefined"`. Press the copy button on the "ID Only" story for this component and it will paste `"undefined"`.

This is also assuming that we don't want to copy functionality for the ID, because it's so short and not truncated anyway. Otherwise I'd probably change this whole part to:

```
{!hideCopy && (address || id) && (
  <CopyText text={address || id} hideCopyText={hideCopyText} color={color} />
)}
```

src/components/AddressLink/index.tsx
navFooh: Seems about right, except that it's `flex-direction`. However `row` is the default value so that's why it still works.

Also I realized that the line-height doesn't need to be conditional here, it can always be `1.5`

src/components/HistoryTables/MessageHistory/Detail.tsx
```diff
@@ -52,14 +50,32 @@ export default function MessageDetail(props: MessageDetailProps) {
...
+    const baseFeeBurn = new FilecoinNumber(
+      (message as Message)?.gasCost?.baseFeeBurn,
+      'attofil'
+    )
+    const overEstimationBurn = new FilecoinNumber(
+      (message as Message)?.gasCost?.overEstimationBurn,
+      'attofil'
+    )
+    return formatNumber(baseFeeBurn.plus(overEstimationBurn).toString())
+  }, [message, pending])
```
navFooh: Here optional chaining is used to access `baseFeeBurn` and `overEstimationBurn`, but it seems like it is assumed that the values do exist, otherwise `undefined` will be passed to the `FilecoinNumber` constructor. I assume this would result in a value of `0` though, but not sure if intentional.

Then finally `toString` is used instead of `toFil` or `toAttoFil`. It should be equivalent to `toFil`, but since we're dealing with Filecoin values, I think it'd be best to be specific?

src/components/HistoryTables/useAllMessages/index.ts
```diff
@@ -340,13 +340,10 @@ export const useMessage = (cid: string, height?: number) => {
...
-    const msg = message as MessageConfirmed
-    const bnBaseFeeBurn = new BigNumber(msg?.baseFeeBurn)
-    const bnGasBurned = new BigNumber(msg?.gasBurned)
+    const msg = message as Message
+    const totalCost = new BigNumber(msg?.gasCost?.totalCost)
 
-    return (
-      !!msg && !bnBaseFeeBurn.isGreaterThan(0) && !bnGasBurned.isGreaterThan(0)
-    )
+    return !!msg && !totalCost
```
navFooh: I'm not sure about the check here in the end. Won't `!totalCost` always result in `false` because this is an object, a `BigNumber`? Shouldn't we check `!totalCost.isGreaterThan(0)` to keep similar functionality as before?

src/components/HistoryTables/utils/index.ts
```diff
@@ -1,50 +1,18 @@
...
-  if (pending) return ''
-  message = message as MessageConfirmed
-  const bnBaseFeeBurn = new BigNumber(message.baseFeeBurn)
-  const bnOverEstimationBurn = new BigNumber(message.overEstimationBurn)
-  const bnMinerTip = new BigNumber(message.minerTip)
-  const totalCost = bnBaseFeeBurn.plus(bnOverEstimationBurn).plus(bnMinerTip)
-  const totalFil = new FilecoinNumber(totalCost, 'attofil')
-  return makeFriendlyBalance(totalFil) + ' FIL'
-}
-
```
navFooh, line 39: Niiiice!

navFooh (changes requested):

.storybook/main.js
navFooh: sounds good!

src/components/AddressLink/index.tsx
navFooh: This is resolved in https://github.com/glifio/react-components/pull/440

src/components/AddressLink/index.tsx
navFooh: This is resolved in https://github.com/glifio/react-components/pull/440

package.json
Schwartz10: Nope thats a mistake

src/components/HistoryTables/MessageHistory/Detail.tsx
Schwartz10: Yes good catch here

src/components/HistoryTables/useAllMessages/index.ts
Schwartz10: Also v good catch

Schwartz10: https://github.com/glifio/react-components/pull/439/files#diff-b7cf153627a387ac5359abd7f2187f87f056a48fceb205002d3aa718657db150R349 fix

PR #445, opened 2022-07-02
Fixes

Schwartz10:


Schwartz10:
Hey @navFooh i needed to merge this to get the safe working again on mainnet, but would appreciate a review after the fact.

the new backend no longer returns an `id` address when you lookup by `id` address on the Address query. This broke things like (1) checking if an address is a signer of a multisig and (2) checking if the current user's wallet as approved a transaction already 

To fix this, i changed the `Wallet` type to include a `robust` and a `id` key to represent both address types. This allows us to make comparisons based on id address more easily elsewhere across apps. We can phase out the `address` key from the wallet and safe, and remove it from the `Wallet` type.

__mocks__/@glif/filecoin-wallet-provider.js
```diff
@@ -4,14 +4,16 @@ import { Message } from '@glif/filecoin-message'
...
-  const accounts = []
-  for (let i = start; i < end; i++) {
-    accounts.push(`t1mbk7q6gm4rjlndfqw6f2vkfgqotres3fgicb${i}uq`)
-  }
+export const mockGetAccounts = jest
+  .fn()
+  .mockImplementation((start = 0, end = 1) => {
+    const accounts = []
+    for (let i = start; i < end; i++) {
+      accounts.push(`t1mbk7q6gm4rjlndfqw6f2vkfgqotres3fgicb${i}uq`)
```
navFooh, line 12: I imagine the address will be invalid when end > 10, not sure if we should throw an error in that case

PR #456, opened 2022-07-05
Optimizations

Schwartz10:
@navFooh i'm going to get some tests written for this, which i think should basically just make sure things look right in different states and prevent any major regressions.

The point of this PR is to decouple some queries that can take longer to load, to prevent the entire message itself from taking a long time to load. Things like the message's receipt and the gas costs of the message (besides those parameters like limit, premium, and feecap that are included in the message itself) can occasionally take a while to compute. Once this PR is merged ,those queries are non blocking on displaying some basic message details.

This PR will require a follow on in the safe too since the Message no longer has a receipt


src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+  }, [message, msgRcptQuery?.receipt])
+
+  const messageState = useMemo<MessageState>(() => {
+    if (pending) return MessageState.Pending
+    else if (loading) return MessageState.Loading
+    else if (!message && !pending && !error) return MessageState.NotFound
+    else if (!message && !pending && !!error) return MessageState.Error
+    else if (!!message && !gasUsed) return MessageState.Confirmed
+    else if (!!message && !!gasUsed) return MessageState.Executed
+    else return MessageState.Loading
```
Schwartz10, line 88: Still need to think about this a bit more and notice if we see any strange behavior from loading state being the default

src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+            loading={false}
+            error={error}
+          />
+        )}
+        {messageState === MessageState.NotFound && (
+          <StandardBox>
+            <p>Message {cid} not found.</p>
+            <p>
+              Note - it may take 1-2 minutes for a recently confirmed message to
+              show up here.
```
Schwartz10, line 113: We could show a "our database is still syncing mainnet" thing here too, but i couldnt think of a good way to word / design this so i left it for now

src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+            </p>
+          </StandardBox>
+        )}
+        {messageState === MessageState.Loading && (
+          <StandardBox>
+            <LoaderGlyph />
+            <p>Searching for {cid}... Give us a moment</p>
```
Schwartz10, line 120: Thoughts on this UI? I'm not crazy about the loading animation, but (i think) we're still using it elsewhere ,so best to be consistent with loading indication throughout the apps?

src/customPropTypes.ts
```diff
@@ -128,6 +146,16 @@ export const GAS_PARAMS_PROPTYPE = shape({
...
 
+export const GAS_COST_QUERY_PROPTYPE = shape({
+  baseFeeBurn: string.isRequired,
+  minerPenalty: string.isRequired,
+  minerTip: string.isRequired,
+  overEstimationBurn: string.isRequired,
+  refund: string.isRequired,
+  totalCost: string.isRequired,
+  gasUsed: number.isRequired
+})
```
Schwartz10: Originally had this merged wit hthe other Message proptype declared right above it, but it felt more useful to separate them into two different proptype declarations

Schwartz10:
@navFooh i can't figure out why `params` as a key still renders lowercase in this PR https://github.com/glifio/react-components/pull/456/files#diff-ad830a613958d0286523ccd8a0fcf877b44cff4a25f2793b8e0b652999f91e77R1008

Reading the code.. it should say "Parameters"? Let me know if something sticks out at you, if not i can look closer

src/components/HistoryTables/detail.tsx
```diff
@@ -244,7 +258,6 @@ export const Parameters = ({ params, depth, actorName }: ParametersProps) => (
                     />
                   </Lines>
                 )
-              break
```
navFooh: @Schwartz10 The `break` here was the fix for rendering `null` as `-` in the parameters, as otherwise it acts like a fall-through switch statement, rendering `null` as `false`, which was a mistake. I think this accidentally got removed in the merge?

src/components/HistoryTables/MessageHistory/Detail/index.test.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+const expectElementDNE = (text: string) => {
+  try {
+    screen.getByText(text, { exact: false })
+  } catch (err) {
+    expect(
+      err.message.includes(`Unable to find an element with the text: ${text}.`)
+    ).toBeTruthy()
+  }
+}
```
navFooh: This doesn't seem very useful, because it will also succeed if the element _does_ exist, right?
The proper way to test if something does not exist seems to be: `expect(screen.queryByText(/Speed up/)).toBeNull()`
Because `queryByText` doesn't fail, but return `null` if nothing is found.

src/components/HistoryTables/MessageHistory/Detail/index.test.tsx
```diff
@@ -0,0 +1,333 @@
...
+          </ThemeProvider>
+        </MockedProvider>
+      )
+      container = res.container
+    })
+
+    expect(
+      screen.getByText(`Searching for ${cid}... Give us a moment`)
+    ).toBeInTheDocument()
+    // couldnt figure out a better way to assert the speed up and cancel buttons are not display in loading state, but this is not great if we ever change names of Cancel and Speed up ButtonV2Links
```
navFooh: As long as we check that Speed Up and Cancel _do_ exist somewhere else, we should be safeguarded in case the button text changes. It's probably best to reuse the button names from a variable in this test then.

src/customPropTypes.ts
```diff
@@ -112,6 +112,24 @@ export const MESSAGE_PROPTYPE = shape({
...
+  gasLimit: number.isRequired,
+  params: oneOfType([string, arrayOf(string)])
+})
+
+export const MESSAGE_RECEIPT_PROPTYPE = shape({
+  exitCode: number.isRequired,
+  return: string.isRequired,
+  gasUsed: number.isRequired
+})
+
```
navFooh: This makes sense, but I think I'd rather keep the naming similar to what we did with Address:
`ADDRESS_PROPTYPE` / `GRAPHQL_ADDRESS_PROP_TYPE`
`MESSAGE_PROPTYPE` / `GRAPHQL_MESSAGE_PROP_TYPE`

src/components/HistoryTables/MessageHistory/History/index.tsx
```diff
@@ -27,9 +30,21 @@ export default function MessageHistoryTable(props: MessageHistoryTableProps) {
...
+  const daysLeft = useMemo(() => FINISH_DATE.fromNow(true), [])
+
   return (
     <Box>
       <Title>Transaction History</Title>
+      {props.warnMissingData && (
+        <WarningBox>
+          <h3>Limited Historical Data</h3>
+          <p>
+            Some transactions are currently missing for the next {daysLeft}{' '}
```
navFooh, line 42: This won't work really well when we pass the `FINISH_DATE` and `props.warnMissingData` is still true. It will basically start counting up again, because then it will be `x days ago` instead of `in x days`, although it omits `in` / `ago`.

navFooh:
> @navFooh i can't figure out why `params` as a key still renders lowercase in this PR https://github.com/glifio/react-components/pull/456/files#diff-ad830a613958d0286523ccd8a0fcf877b44cff4a25f2793b8e0b652999f91e77R1008
> 
> Reading the code.. it should say "Parameters"? Let me know if something sticks out at you, if not i can look closer

Yeah I see, this happens when parameters is `null`. Maybe this isn't the nicest solution altogether, but that's something we can maybe address when tackling the explorer

src/components/HistoryTables/MessageHistory/Detail/index.tsx
navFooh: Yeah it doesn't look too great.. but for now it's maybe best to keep consistent indeed

navFooh (changes requested):

src/components/HistoryTables/detail.tsx
Schwartz10: Oh shoot this is my mistake, i'll remove

src/components/HistoryTables/MessageHistory/History/index.tsx
Schwartz10: Going to fix this in a follow on PR

PR #457, opened 2022-07-05
Updates to address checks

### LINKED ISSUE #455: Empty params displays "false" on msg detail
image

If params are empty, it should just be an empty string

navFooh:


__mocks__/@glif/filecoin-wallet-provider.js
```diff
@@ -7,11 +7,12 @@ export * from '../../node_modules/@glif/filecoin-wallet-provider/dist/providers/
 export const mockGetAccounts = jest
   .fn()
   .mockImplementation((start = 0, end = 1) => {
+    if (end > 10)
+      throw new Error('Cannot generate a valid account address with an index of 10 or higher')
```
navFooh, line 11: When end > 10 this will not generate valid addresses

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -80,24 +80,18 @@ export default function ProposalDetail(props: ProposalDetailProps) {
     return _msigTxsError
   }, [_msigTxsError, msigTxsLoading, proposal])
 
-  const actionRequired = useMemo(() => {
-    if (proposalFoundError) return false
-    if (!props.walletAddress?.robust && !props.walletAddress?.id) return false
-    if (!!proposal && !stateLoading && !stateError) {
-      const walletAddressIsSigner = isAddressSigner(
-        { id: props.walletAddress?.robust, robust: props.walletAddress?.id },
```
navFooh, line 88: Not sure why this was setting `{ id: ...robust, robust: ...id }`? But should be fine to shorten to `props.walletAddress`

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -80,24 +80,18 @@ export default function ProposalDetail(props: ProposalDetailProps) {
     return _msigTxsError
   }, [_msigTxsError, msigTxsLoading, proposal])
 
-  const actionRequired = useMemo(() => {
-    if (proposalFoundError) return false
-    if (!props.walletAddress?.robust && !props.walletAddress?.id) return false
-    if (!!proposal && !stateLoading && !stateError) {
```
navFooh, line 86: All cases for returning false can be merged in one check, if `props.walletAddress` has no `id` or `robust`, `isAddressSigner` will already return false

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -80,24 +80,18 @@ export default function ProposalDetail(props: ProposalDetailProps) {
...
+    if (!proposal || proposalFoundError || stateError || stateLoading)
+      return false
+    const walletAddressIsSigner = isAddressSigner(
+      props.walletAddress,
+      stateData.State.Signers
+    )
+    const alreadyApproved = isAddressSigner(
+      props.walletAddress,
+      proposal.approved
+    )
```
navFooh, line 93: Even though the naming of the method is not perfect for here, what it does is the same

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -107,19 +101,10 @@ export default function ProposalDetail(props: ProposalDetailProps) {
     stateError
   ])
 
-  const isProposer = useMemo(() => {
-    if (
-      !proposalFoundError &&
-      !!proposal &&
-      !!(props.walletAddress?.id || props?.walletAddress?.robust)
```
navFooh, line 114: I believe all these checks are already covered by `isAddrEqual` returning false in these cases

src/components/HistoryTables/MsigProposals/index.tsx
```diff
@@ -43,9 +43,7 @@ export default function ProposalHistoryTable(props: ProposalHistoryTableProps) {
                   idHref={props.idHref}
                   inspectingAddress={props.address}
                   actionRequired={
-                    !proposal.approved.some(address =>
-                      isAddrEqual(address, props.walletAddr)
-                    )
+                    !isAddressSigner(props.walletAddr, proposal.approved)
```
navFooh, line 46: Again, method name is not ideal, but does the same

src/components/Layout/Lines.tsx
```diff
@@ -50,12 +51,9 @@ const LineEl = styled.div`
   }
 `
 
-const capitalFirstLetter = (word: string): string =>
-  word.slice(0, 1).toUpperCase() + word.slice(1).toLowerCase()
-
 export const Line = ({ label, depth, children }: LineProps) => (
   <LineEl depth={depth}>
-    <div>{capitalFirstLetter(label)}</div>
+    <div>{label}</div>
```
navFooh, line 56: The CSS rule will capitalize each word, as opposed to only the first letter in the string, but might be prefeble and lighter than using Javascript

src/utils/isAddrEqual/index.test.ts
```diff
@@ -28,6 +37,21 @@ describe('isAddrEqual', () => {
...
+        { id: WALLET_ID, robust: '' },
+        { id: WALLET_ID_2, robust: '' }
+      )
+    ).toBe(false)
+    expect(
+      isAddrEqual(
+        { id: '', robust: WALLET_ADDRESS },
+        { id: '', robust: WALLET_ADDRESS_2 }
+      )
+    ).toBe(false)
```
navFooh, line 51: These cases were falsely returning `true` before. If both arguments were of the `Address` type and both the `id` or `robust` values were empty, it would match `'' === ''` as true.

I think maybe we should safeguard for this, which the updated `isAddrEqual` does.

src/utils/isAddrEqual/index.ts
```diff
@@ -7,15 +7,14 @@ export const isAddrEqual = (
...
-    convertAddrToPrefix(address2.id) === convertAddrToPrefix(address.id) ||
-    convertAddrToPrefix(address2.robust) === convertAddrToPrefix(address.robust)
+  const id1 = convertAddrToPrefix(address.id)
+  const robust1 = convertAddrToPrefix(address.robust)
+  const id2 = convertAddrToPrefix(
+    typeof address2 === 'string' ? address2 : address2.id
+  )
+  const robust2 = convertAddrToPrefix(
+    typeof address2 === 'string' ? address2 : address2.robust
   )
```
navFooh, line 17: This replaces the previous logic for using a string for `address2`, allowing a single return statement

src/utils/isAddrEqual/index.ts
```diff
@@ -7,15 +7,14 @@ export const isAddrEqual = (
...
+
+  return (!!id1 && id1 === id2) || (!!robust1 && robust1 === robust2)
```
navFooh, line 19: This prevents returning `true` for:
```
isAddrEqual(
  { id: '', robust: WALLET_ADDRESS },
  { id: '', robust: WALLET_ADDRESS_2 }
)
```

src/utils/isAddressSigner/index.ts
```diff
@@ -4,8 +4,4 @@ import { isAddrEqual } from '../isAddrEqual'
 export const isAddressSigner = (
   address: Address,
   signers: Address[]
-): boolean =>
-  signers.some(
-    signer =>
-      isAddrEqual(signer, address?.robust) || isAddrEqual(signer, address?.id)
-  )
+): boolean => signers.some(signer => isAddrEqual(signer, address))
```
navFooh, line 7: Since `isAddrEqual` can test two `Address` ojects, this can be simplified

src/utils/isAddressSigner/index.ts
```diff
@@ -4,8 +4,4 @@ import { isAddrEqual } from '../isAddrEqual'
 export const isAddressSigner = (
   address: Address,
   signers: Address[]
```
navFooh, line 6: We could potentially rename this whole method to `isAddressPresent(address: Address, addresses: Address[])` because it dosn't necessarily have to be used to find a signer

src/components/HistoryTables/MsigProposals/Proposal.tsx
Schwartz10: Wow.. good catch

src/utils/isAddrEqual/index.test.ts
Schwartz10: Good catch!

Schwartz10 (approved):

PR #465, opened 2022-07-06
Update message detail

navFooh:


src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -73,25 +74,28 @@ export default function MessageDetail(props: MessageDetailProps) {
...
-    else return MessageState.Loading
+    if (!message) return MessageState.NotFound
+    if (!gasUsed) return MessageState.Confirmed
+    return MessageState.Executed
   }, [pending, message, error, gasUsed, loading])
 
+  // Log errors
+  useEffect(() => error && logger.error(error), [error])
+  useEffect(() => gasStateError && logger.error(gasStateError), [gasStateError])
+  useEffect(() => msgRcptError && logger.error(msgRcptError), [msgRcptError])
```
navFooh, line 97: Logs any errors we're getting from GraphQL to Sentry and the console

src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -73,25 +74,28 @@ export default function MessageDetail(props: MessageDetailProps) {
...
     if (pending) return MessageState.Pending
-    else if (loading) return MessageState.Loading
-    else if (!message && !pending && !error) return MessageState.NotFound
-    else if (!message && !pending && !!error) return MessageState.Error
-    else if (!!message && !gasUsed) return MessageState.Confirmed
-    else if (!!message && !!gasUsed) return MessageState.Executed
-    else return MessageState.Loading
+    if (!message) return MessageState.NotFound
+    if (!gasUsed) return MessageState.Confirmed
+    return MessageState.Executed
```
navFooh, line 91: I'd simplify this a bit:
- because we `return` after `if`, there is no need to use `else`
- if, for example, we return on `if (pending)`, we never need to check for `if (!pending)` after, pending will always be falsy
- Keeping the above in mind, we can change the order so that one check per line covers all the cases. On the final return:
  - we have no error
  - we're not loading
  - we're not pending
  - we have a message
  - we have gas parameters
  
  meaning the message was successfully executed

src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -73,25 +74,28 @@ export default function MessageDetail(props: MessageDetailProps) {
...
-    if (toExec && receiptExists && initTx) {
-      return getAddrFromReceipt(msgRcptQuery?.receipt?.return)
-    }
-    return null
+    const isToExec = isAddrEqual(message?.to, 'f01')
+    const isInitTx = Number(message?.method) === 2
+    const receiptReturn = msgRcptQuery?.receipt?.return
+    return isToExec && isInitTx && receiptReturn
+      ? getAddrFromReceipt(receiptReturn)
+      : null
```
navFooh, line 82: just renamed the variables a bit and stored `msgRcptQuery?.receipt?.return` because we can just check if it's truthy and reuse it

src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -47,17 +48,17 @@ export default function MessageDetail(props: MessageDetailProps) {
...
     variables: { cid },
     pollInterval: 10000
   })
 
   const transactionFee = useMemo<string>(() => {
-    if (pending || !gasQuery?.gascost?.totalCost) return 'Pending...'
+    if (pending) return 'Pending...'
     const cost = gasQuery?.gascost?.totalCost
     return cost
       ? `${makeFriendlyBalance(new FilecoinNumber(cost, 'attofil'))} FIL`
```
navFooh, line 64: If you return `'Pending...'` on `!gasQuery?.gascost?.totalCost`, the second part of the ternary statement (where it returns `''`) will never be hit. If we do want to return `'Pending...'` when cost is not yet known, we should do this:

```javascript
const cost = gasQuery?.gascost?.totalCost
return cost && !pending
  ? `${makeFriendlyBalance(new FilecoinNumber(cost, 'attofil'))} FIL`
  :  'Pending...'
```

navFooh: @Schwartz10 let me know which you prefer

Schwartz10: I prefer not _only_ relying on `pending` as a prop to determine if `gasQuery?.gascost?.totalCost` exists, since there is a state where the message is not pending, but the gascost has still yet to be computed. So i'd go with the version in the comment rather than the one commited

src/components/HistoryTables/MessageHistory/Detail/index.tsx
Schwartz10: I prefer a more explicit approach to determining if a message was executed rather than this more "implied" approach.

Reason being is messy and not necessarily "good", but I have a feeling there could be a state where:
```
we have no error
we're not loading
we're not pending
we have a message
we have gas parameters
```

Yet still, for whatever reason, the message's gas costs and receipt have not been populated for one reason or another

src/components/HistoryTables/MessageHistory/Detail/index.tsx
navFooh: It's not really a matter of relying on pending as a prop to determine if `gasQuery?.gascost?.totalCost` exists.

The question is more what we want to show in either case.

In the previous version, it basically said:
- Show "Pending..." if `gasQuery?.gascost?.totalCost` doesn't exist or we are pending
- And then show "" if `gasQuery?.gascost?.totalCost` doesn't exist

The latter would never be hit due to the previous statement.

src/components/HistoryTables/MessageHistory/Detail/index.tsx
navFooh: That I understand, but I wonder what the difference is with the previous approach, as it was also checking the same conditions, but in a different syntax?

I was also wondering if we should check te receipt for determining if the message was actually executed

Schwartz10 (approved):

PR #466, opened 2022-07-08
Fix select issue

navFooh:
See comments below

### DIFF
--- a/src/components/InputV2/Select.tsx
+++ b/src/components/InputV2/Select.tsx
@@ -51,14 +51,12 @@ export const Select = ({
   onEnter,
   setIsValid
 }: SelectProps) => {
-  // Make sure value is always a valid option or
-  // an empty string when there is a placeholder
-  useEffect(() => {
-    if (!options.includes(value)) {
-      const defaultValue = placeholder ? '' : options[0]
-      if (value !== defaultValue) onChange(defaultValue)
-    }
-  }, [value, placeholder, options, onChange])
+  // Automatically select the first option if value
+  // is not set and there is no placeholder text
+  useEffect(
+    () => !value && !placeholder && options[0] && onChange(options[0]),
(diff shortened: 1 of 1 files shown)


src/components/InputV2/Select.tsx
```diff
@@ -51,19 +51,17 @@ export const Select = ({
...
-      const defaultValue = placeholder ? '' : options[0]
-      if (value !== defaultValue) onChange(defaultValue)
-    }
-  }, [value, placeholder, options, onChange])
+  // Automatically select the first option if value
+  // is not set and there is no placeholder text
+  useEffect(
+    () => !value && !placeholder && options[0] && onChange(options[0]),
+    [value, placeholder, options, onChange]
+  )
```
navFooh, line 59: @Schwartz10 The idea here was to ensure that `value` is always a valid `option`. However, this works against us as sometimes we already know which value we want, but options are still being populated (in the case of change / remove signer).

What we still want to happen though, is that we automatically select the first available option if `value` is not set and there is no `placeholder` text. This allows to pass an array of `options` but leave `value` empty and let the component automatically select the first option, which would be expected to show as selected in the UI as well.

src/components/InputV2/Select.tsx
```diff
@@ -51,19 +51,17 @@ export const Select = ({
...
 
   // Communicate validity to parent component
   useEffect(
-    () => setIsValid(!required || !!value),
-    [value, required, setIsValid]
+    () => setIsValid(value ? options.includes(value) : !required),
+    [value, options, required, setIsValid]
```
navFooh, line 64: @Schwartz10 This will flag the input as invalid if value is not present in the options, which can happen when an option is selected but the list of options is still populating

Schwartz10 (approved):

PR #467, opened 2022-07-08
Add sync status to message history

Schwartz10:


src/components/SyncStatus/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  useChainHeadSubscription({
+    variables: {},
+    shouldResubscribe: chainHead === 0,
+    onSubscriptionData: ({ subscriptionData }) => {
+      if (
+        !subscriptionData.loading &&
+        !subscriptionData.error &&
+        chainHead === 0
+      ) {
+        setChainHead(subscriptionData.data?.chainHead?.height as number)
```
navFooh: `subscriptionData.data?.chainHead?.height` should already be correctly interpreted as `number`. I think it's best not to overwrite the type derived from the GraphQL schema, so TypeScript will correctly warn us of any changes.

src/components/SyncStatus/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+dayjs.extend(relativeTime.default)
+
+export const SyncStatus = () => {
+  const [chainHead, setChainHead] = useState(0)
+  const { data } = useStatusQuery()
+
+  useChainHeadSubscription({
+    variables: {},
+    shouldResubscribe: chainHead === 0,
```
navFooh, line 19: Why do we need to resubscribe here?

src/components/SyncStatus/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  const syncPercentage = useMemo(() => {
+    if (data?.status?.height > 0 && chainHead > 0) {
+      return new BigNumber((chainHead - data?.status?.height) / chainHead)
+        .multipliedBy(100)
+        .toFixed(0)
+    }
+    return 0
+  }, [data?.status?.height, chainHead])
```
navFooh: Both `data?.status?.height` and `chainHead` are of type `number`, so there should be no reason to convert to `BigNumber` in this calculation. Also, `toFixed()` returns a string and the default return value is a number, which is not the best practice. That's why I'd prefer to always add the generic type to `useMemo`, so typescript will catch inconsistencies or at least make it obvious when there are different return types (`<string | number>`).

src/components/SyncStatus/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    if (data?.status?.estimate > 0) {
+      return dayjs().add(data.status.estimate, 'seconds').fromNow()
+    }
+    return ''
+  }, [data?.status?.estimate])
+
+  return (
+    <WarningBox>
+      <h3>Sync in Progress</h3>
+      {syncPercentage > 0 && (
```
navFooh, line 52: Checking `syncPercentage  > 0` is not the best since `syncPercentage` can be a string

src/components/SyncStatus/index.tsx
navFooh: I'd rather do something like `Math.round((chainHead - data?.status?.height) / chainHead * 100)` to get an even number, potentially wrap it in `Math.min(100, Math.max(0, ...))` to ensure it's never less than `0` or more than `100`.

navFooh (changes requested):

src/components/SyncStatus/index.tsx
Schwartz10: The server can sometimes drop connection with the filecoin node, which then drops the connection from the web app client to the server. If that happens and we haven't got the data, we should resub

navFooh: I see, that makes sense.

Reading the documentation for `shouldResubscribe`, it doesn't make it obvious to me that this option works that way:
> Determines if your subscription should be unsubscribed and subscribed again

This rather tells me, "when this value is / becomes truthy, stop and restart the connection", not "when the connection drops, resubscribe if this value is truthy".

The latter should happen by default I imagine. Though if you're sure that this works as intended, it's good for me

Schwartz10: Ah yeah maybe i'm misunderstanding... The documentation for `shouldResubscribe` isn't super helpful, but i found this comment https://github.com/apollographql/apollo-client/issues/7511#issuecomment-768093125 which (could be wrong) says that shouldResubscribe is default true. If anything, this at least says _dont_ resubscribe (for whatever reason we were unsubbed in the first place..) if we already have the data.

Schwartz10: I'll keep it for now and if we notice bandwidth going crazy on the backend or something, we can try disabling

src/components/SyncStatus/index.tsx
Schwartz10: After the changes, doesn't seem like syncPercentage will ever be a string, so will keep this for now

src/components/SyncStatus/index.tsx
navFooh: sounds good!

navFooh (approved):

PR #479, opened 2022-07-21
Feat/refactor account selector

Schwartz10:
Things left to do:
- [x] Handle "Selected" UI 
- [x] Write new tests
- [x] Remove unused code


Schwartz10:
@navFooh tests are passing for me locally, but will fail until we get the graph-calibration back up and running

Also during your review, would definitely appreciate your thoughts on the UI / spacing of things.

PR #481, opened 2022-07-27
Layout updates

navFooh:
- Use space variables from base-css instead of theme
- Created `LabeledText` component
- Derive `AddressLink` from `LabeledText` and move to `LabeledText` folder
- Removed `AppIconWrapper` since we're not using the App title in the header
- Created `SearchBar` component and `SearchAddressMessage` which derives from `SearchBar`
- Allowed possibility of custom components in `AppHeader` (like `SearchBar`)
- Allowed `LabeledText` components in `AppHeader`
- Removed back button from `AppHeader` since we're not using it
- Created `PageTitle` component in `Layout` to reuse instead of `Title` from `HistoryTables/generic`
- Created `Badge` component without `styled-system` and moved to `Layout`, added stories
- Updated `OneColumn` spacing to improve layout in explorer


Schwartz10 (changes requested):
this mostly looks good to me, a couple comments:

- Should we stick to the same naming standard for size/color so like `space-medium` instead of `space-m`?
- In the search bar, when you hover the x button, `cursor: pointer`
- When embedding the search bar in the AppHeader, can we extend the search bar's width? it should be able to fit a full CID imo. When the screen gets narrower, the search bar can narrow proportionately with the view size

Also, if you pull primary you can get tests passing.

navFooh:
> * Should we stick to the same naming standard for size/color so like `space-medium` instead of `space-m`?

For sizes I think that abbreviations work a bit better, as they're intuitive and variable names can get quite long otherwise, thinking about `space-extra-large`, `space-negative-medium`, etc. Since we don't have code completion on CSS variables they're a bit more prone to typos as well. Also for colors I think that abbreviations don't work that well.

> * In the search bar, when you hover the x button, `cursor: pointer`

Done through `base-css`.

> * When embedding the search bar in the AppHeader, can we extend the search bar's width? it should be able to fit a full CID imo. When the screen gets narrower, the search bar can narrow proportionately with the view size

Fixed as well. This required a slight change in the AppHeader behaviour, removing the flex wrap on the left and right parts of the AppHeader. But since we don't necessarily design for small screens, this should be fine.

PR #492, opened 2022-07-29
Table styling updates

navFooh:


Schwartz10 (commented):
Looks good here. A couple of side notes:

- In some circumstances, the colored buttons are a little too eye catching imo. I made this ticket, but we can follow up with that at a later time https://github.com/glifio/react-components/issues/497
- I fixed an issue where clicking the copy button caused the account to get selected

PR #493, opened 2022-07-29
Polish

### LINKED ISSUE #461: Remove missing data WarningBox and make a smaller, nicer less conspicuous warning instead
Just making an issue for this so we don't forget.

The `daysLeft` indication at the following text will start counting back up after `FINISH_DATE` has passed:

```
Some transactions are currently missing for the next {daysLeft}{' '}
while our servers finish syncing with Filecoin Mainnet.
```

Schwartz10:
Add some tests

Also removed the larger sync status warning in favor of something less intrusive and obstructive


navFooh:
Hey, I saw that it's a recurring pattern that we add content next to the page title so I updated to `PageTitle` component to accomodate this. It also makes sure that the margins of the `h2` element are applied to the wrapper so that collapsing margins still work properly with content below.

I also noticed that the column count wasn't consistent throughout the proposal table due to conditionally adding an "Action required" cell. I made some adjustments so that the whole column is omitted when none of the proposals have an "Action required".

Also added a missing `hr` below the page title in the message detail page.

src/components/HistoryTables/MsigProposals/ProposalRow.tsx
```diff
@@ -67,14 +73,16 @@ type ProposalHistoryRowProps = {
   proposal: MsigTransaction
   idHref: (id: number) => string
   inspectingAddress?: string
-  actionRequired?: boolean
+  proposalActionRequired?: boolean
+  anyProposalActionRequired?: boolean
```
Schwartz10: What is the difference between these two?

src/components/HistoryTables/MsigProposals/ProposalRowColumnTitles.tsx
```diff
@@ -1,14 +1,24 @@
...
+  anyProposalActionRequired
+}: ProposalRowColumnTitlesProps) => (
+  <thead>
+    <tr>
+      <th>ID</th>
+      <th>Method</th>
+      <th>Proposer</th>
+      <th>Value</th>
+      <th># Approvals</th>
+      {anyProposalActionRequired && <th />}
```
Schwartz10: When would this be false?

src/components/HistoryTables/MsigProposals/index.tsx
```diff
@@ -7,19 +8,43 @@ import {
...
 
+  const proposals = useMemo<Array<MsigTransaction>>(
+    () =>
+      data?.msigPending
+        ? [...data?.msigPending].sort((a, b) => a.id - b.id)
+        : [],
+    [data?.msigPending]
+  )
+
+  const anyProposalActionRequired = useMemo<boolean>(
```
Schwartz10: Still slightly confused at this

Schwartz10 (commented):
This mostly looks good, just not sure about the change in props for Proposal table

PR #514, opened 2022-08-16
Temp fix for #503

### LINKED ISSUE #503: Transaction.Form should widen to show the entire Input.Address address text
image

It's important that a user can see the full address to verify what they're doing in the transaction

Schwartz10:
@navFooh this temporarily fixes #503 but it feels like the "right" way to do this is to set XXX:

```
          > *:last-child:not(.toggle-wrapper) {
            flex: 0 0 XXX%;
          }
```

based on the content the `Label` is rendering, and set a `max-width` on the address input large enough to fit the full robust address text. If there are no `InputV2.Address` elements, we can set this param lower (like 40-50%). That way if you're showing a form UI with just `InputV2.Number`s (or other smaller inputs), the form UI doesn't look so lopsided (it looks fine when there is an address above it). 

Wasn't sure if you'd agree with this idea or the approach, interested in your feedback.


navFooh:
Yeah it's very hard to do this "right", as in dynamically making the right side of the labels larger depending on whether there is an address input present in the form. Because they are independent flex rows, the row below has no notion of the content in the row above. This would be different in a HTML table, but that would be quite a big haulover.

We also wouldn't want to pass a property to all of the `InputV2` components just for the sake of controlling that width.

Maybe a better approach would then be to add some styling to the containing element. I can have a look at that.

Also I think we should give the left side of the label a bit more space, as it now wraps the label text and the error message to the next line:

image

navFooh:
I have a slightly modified proposal for now. What if we make the Dialog just a bit wider and reduce the proportional width of the input field a bit? It would look like this:

image

Schwartz10:
> I have a slightly modified proposal for now. What if we make the Dialog just a bit wider and reduce the proportional width of the input field a bit? It would look like this:

@navFooh i like this approach a lot - would you mind making the commit for this?

navFooh:
Done. Also added a story for InputV2 where it shows multiple inputs in a form and tweaked the other InputV2 stories so that they are rendered more accurately to how we use them in the apps. That helps with checking the propertions of the text and input fields

PR #519, opened 2022-08-20
Feat/env context

### LINKED ISSUE #452: Footer links don't resolve on mainnet?

### LINKED ISSUE #490: Move env vars into a context provider

Schwartz10:


src/components/HistoryTables/defaultCacheConfig.ts
```diff
@@ -16,8 +15,6 @@ const parseParams = (_: any, incoming: any) => {
     }
     return null
   } catch (e) {
-    const msg = e instanceof Error ? e.message : String(e)
```
Schwartz10, line 19: @navFooh i couldn't think of a great way to replicate the logger here... i'm not convinced we need to log errors here though?

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export type NetworkInfo = {
+  nodeStatusApiKey: string
+  graphUrl: string
+  lotusApiUrl: string
+  networkName: Network
+  coinType: CoinType
+}
+
+export interface Environment {
```
navFooh: There are two exports callled `Environment`, both this interface and the React wrapper component below. Typescript doesn't seem to complain about this, but I'm wondering if it's good practice?

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  [Network.WALLABY]: {
+    nodeStatusApiKey: '',
+    graphUrl: '',
+    lotusApiUrl: '',
+    networkName: Network.WALLABY,
+    coinType: CoinType.TEST
+  }
+}
+
+export const Environment = ({
```
navFooh, line 89: This is the other `Environment` export

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export const useEnvironment = (): Environment => {
+  return useContext(EnvironmentContext)
+}
+
+export const useLogger = (): Logger => {
+  return useContext(EnvironmentContext).logger
+}
+
+export type EnvironmentPropTypes = {
```
navFooh: I'd just call this `EnvironmentProps` as the convention with other react components, where PropTypes refers to an object of React PropTypes for the component.

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+        sentryEnabled: environment.isProd,
+        sentryLevel: LogLevel.WARN,
+        sentryDsn: environment.sentryDsn || '',
+        sentryEnv: environment.sentryEnv || '',
+        sentryTraces: 0,
+        packageName: 'react-components',
+        packageVersion: '?.?.?'
+      }),
+    [environment.isProd, environment.sentryDsn, environment.sentryEnv]
+  )
```
navFooh: Do you think we can keep `logger` as a separate entity and just get the properties from `useEnvironment` there?
Though I do like that it's created through a `useMemo` here so that it reinitializes when something changes. But it also feels weird to have it nested in environment. Just wonder what you think the implications would be to have them separated

src/components/AccountCard/index.test.jsx
```diff
@@ -1,72 +1,110 @@
...
+        networkName={Network.CALIBRATION}
+        nodeStatusApiKey='m787669344-2a9b90eb03dbff3e503c93c7'
+        graphUrl='graph-calibration.glif.link/query'
+        lotusApiUrl='https://api.calibration.node.glif.io/'
+        explorerUrl='https://explorer-calibration.glif.link'
+      >
+        <ThemeProvider theme={theme}>
+          <Base {...Base.args} />
+        </ThemeProvider>
+      </Environment>
```
navFooh: I'm a bit worried about all the repitition of creating this `Environment` wrapper for every test. Here it feels like we're going backwards a bit, because it seems a hassle to maintain if this is the standard for every test that requires something from `Environment`.

Can we create a `TestEnvironment` somewhere that we can just include and use as-is? Even more ideally, all of the apps could also include this same `TestEnvironment` from react-components?

We could use the same for stories, but there it can be a bit simpler even maybe. It seems you can make a global decorator for all of the stories:
https://stackoverflow.com/questions/67748366/how-to-add-a-global-decorator-in-storybook

navFooh:
I wonder, for the Account Creation screen, the user can now choose between coin types "Mainnet" and "Legacy" which map to "t" and "f". There are some methods to convert between one and the other.

Does this also need to be updated to support a new coin type for the Wallaby net?

src/components/AccountSelector/table.tsx
```diff
@@ -1,15 +1,18 @@
...
-import { WALLET_PROPTYPE } from '../../customPropTypes'
+import { COIN_TYPE_PROPTYPE, WALLET_PROPTYPE } from '../../customPropTypes'
+import { useEnvironment } from '../../services/EnvironmentProvider'
 
 export const idxFromPath = (path: string): string => {
   return path.split('/')[5]
 }
 
 const WalletRow = ({
+  coinType,
```
navFooh, line 15: Should we maybe just get the `coinType` from `useEnvironment` in the `WalletRow` instead of passing it as a prop?

src/components/HistoryTables/MessageHistory/hooks/useAllMessages.ts
```diff
@@ -20,12 +22,13 @@ const WAIT_EPOCHS_BEFORE_REFRESH = 3
 
 export const usePendingMessages = (
   address: string,
-  chainHeadSubscription: SubscriptionResult<ChainHeadSubscription, any>
+  chainHeadSubscription: SubscriptionResult<ChainHeadSubscription, any>,
+  coinType: CoinType
```
navFooh, line 26: Can't we get the coinType from `useEnvironment()` like in `useAllMessages`? Just seems a bit cleaner and makes more sense when using `usePendingMessages` on it's own, if that ever happens

src/components/HistoryTables/defaultCacheConfig.ts
navFooh: Well probably this parseParams will not be necessary in the future when implementing the explorer update. Maybe I would leave a `console.error` just as a reference in case we see weird things

src/utils/useGetMessage.ts
```diff
@@ -20,6 +15,15 @@ export const useGetMessage = (cid: string): UseGetMessageResult => {
...
   const [error, setError] = useState<Error | null>(null)
+  const { lotusApiUrl: apiAddress } = useEnvironment()
+
+  const lotusRPC = useMemo(
+    () =>
+      new LotusRPCEngine({
+        apiAddress
+      }),
+    [apiAddress]
+  )
```
navFooh, line 26: Just an idea, should we maybe create a `LotusRPCEngine` instance in the `EnvironmentProvider` when changing the `apiAddress`? Then we can just do something like `const { lotus } = useEnvironment()`

src/test-utils/composeMockAppTree/index.jsx
```diff
@@ -27,14 +29,23 @@ const Index = (statePreset = 'preOnboard', options = {}) => {
...
-        statePreset={statePreset}
-        getState={cacheWalletProviderState}
-        initialState={initialState}
+      <Environment
+        coinType={CoinType.TEST}
+        networkName={Network.CALIBRATION}
+        nodeStatusApiKey='m787669344-2a9b90eb03dbff3e503c93c7'
+        graphUrl='graph-calibration.glif.link/query'
+        lotusApiUrl='https://api.calibration.node.glif.io/'
+        explorerUrl='https://explorer-calibration.glif.link'
```
navFooh: I see now that the Environment is included in the `Tree` here. Can't we better use this in the tests than redefining the `Environment` properties?

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  return <Environment {...env}>{children}</Environment>
+}
+
+export const useEnvironment = (): Environment => {
+  return useContext(EnvironmentContext)
+}
+
+export const useLogger = (): Logger => {
+  return useContext(EnvironmentContext).logger
+}
```
navFooh: If we do keep logger in here, I wouldn't see much harm in using `const { logger } = useEnvironment()` either. It's a bit more transparent as to where it lives. But not really crucial in any case.

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+    networkName: Network.MAINNET,
+    coinType: CoinType.MAIN
+  },
+  [Network.CALIBRATION]: {
+    nodeStatusApiKey: 'm787669344-2a9b90eb03dbff3e503c93c7',
+    graphUrl: 'graph-calibration.glif.link/query',
+    lotusApiUrl: 'https://api.calibration.node.glif.io/',
+    networkName: Network.CALIBRATION,
+    coinType: CoinType.TEST
+  },
```
navFooh, line 79: Don't we want the app urls to update with the environment as well? So that links for the explorer will point to `explorer-calibration.glif.link` when running calibration and to `explorer.glif.io` when running mainnet.

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+      new Logger({
+        consoleEnabled: true,
+        consoleLevel: LogLevel.DEBUG,
+        sentryEnabled: environment.isProd,
+        sentryLevel: LogLevel.WARN,
+        sentryDsn: environment.sentryDsn || '',
+        sentryEnv: environment.sentryEnv || '',
+        sentryTraces: 0,
+        packageName: 'react-components',
+        packageVersion: '?.?.?'
```
navFooh: Should we pass package name and version in the environment as well, so that the apps can set these properties for Sentry?

src/services/EnvironmentProvider/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+export const EnvironmentProvider = ({
+  children,
+  ...initialEnvironment
+}: Omit<
+  EnvironmentPropTypes,
+  'nodeStatusApiKey' | 'graphUrl' | 'lotusApiUrl' | 'coinType' | 'networkName'
+>) => {
+  const router = useRouter()
+  let network = getQueryParam.string(router, 'network') as Network
+  if (!network) network = Network.MAINNET
```
navFooh, line 130: I like the use of the query parameters for this! But won't this also make `app-calibration.glif.link` websites default to mainnet? Not sure if that matters

navFooh (changes requested):

src/services/EnvironmentProvider/index.tsx
Schwartz10: I think the idea here is:

1. For SEO reasons, we want to unify multi network apps into one domain (for all apps)
2. For development reasons, we want a "bleeding edge" deployment 

For this reason, i think keeping the app URLs the same regardless of network makes sense. BUT, realizing we need to add some extra logic to app-to-app navigation that basically pushes a user to explorer.glif.io/?network=calibration when (as one example) the user navigates to the explorer from the wallet when calibration net is the selected network.

src/services/EnvironmentProvider/index.tsx
Schwartz10: The packageName would always be react-components here wouldn't it? How would we get the version? Just right from the package-lock? Wondering if nextjs will let that happen from the parent application?

src/services/EnvironmentProvider/index.tsx
Schwartz10: Yeah i think there's a mindset shift after this PR gets merged - the staging sites (-calibration) no longer determine the network - we can still change the default network on the staging sites to be calibration network by passing the right params from the parent apps (and i'm ok with the staging site always having the `?network=network` query param)

src/test-utils/composeMockAppTree/index.jsx
Schwartz10: Yeah, the `Tree` just renders a lot of stuff in the test environment that isn't always necessary for certain tests. For instance, the CIDLink tests shouldn't need the wallet-provider, balance poller...etc - feels like that could make the tests a lot slower and more clunky

I do agree though, it would be nice to have a standard test environment to standardize all tests across all repos. The current one is pretty messy and could be improved and streamlined. Feels like i could advocate for that being a separate follow on PR in the future though. Since for now, i don't see these changes (albeit a bit repetitive) as _too_ difficult to maintain, since we're not constantly adding new tests to old components? 

Happy to be convinced otherwise though, if you feel strongly we should revise the test env in this PR i can work through it

src/services/EnvironmentProvider/index.tsx
Schwartz10: See https://github.com/glifio/react-components/pull/519/commits/79575566f8e1c5054928e01a585db741471579db

Schwartz10:
> I wonder, for the Account Creation screen, the user can now choose between coin types "Mainnet" and "Legacy" which map to "t" and "f". There are some methods to convert between one and the other. Does this also need to be updated to support a new coin type for the Wallaby net?

Good question and thought! Short answer no, explanation below.

The CoinType is used for replay attack protection on Filecoin mainnet. The CoinTypeCode (`461` (mainnet) or `1` test) is used in `createPath` to derive the private key (and associated public address) of a Filecoin account. We only need to make sure the same mainnet wallet does not transact on _any_ testnet (or else those testnet transactions, despite it being highly improbable, could be replayed on mainnet). We don't care if a wallaby transaction is replayed on calibrationnet tho.

navFooh (approved):

PR #525, opened 2022-08-23
Feat/env context update 2

navFooh:


src/components/SmartLink/index.tsx
```diff
@@ -33,22 +34,34 @@ export const SmartLink = ({
     [href, isMailToOrTelUrl]
   )
 
+  const isGlifUrl = useMemo<boolean>(() => {
+    if (isInternalUrl) return true
+    if (isMailToOrTelUrl) return false
+    const urlObject = new URL(href)
+    const hostNameParts = urlObject.hostname.split('.')
+    const domainName = hostNameParts[hostNameParts.length - 2]
+    return domainName === 'glif'
```
Schwartz10, line 43: Is this the safest way to accomplish getting the hostName? Does the `router` object not give us anything handy here?

PR #530, opened 2022-08-26
Fix issue with env vars not hydrating env

Schwartz10:


Schwartz10:
@navFooh this one's a little less straight forward, but the env vars from the parent app (wallet, safe) weren't getting hydrated in the env provider because of how the initial state was set combined with how the page gets hydrated server side before the env vars are read (client side). This is at least my understanding of what was broken when i began integrating with the wallet.

I'm going to merge this and unblock myself, but would appreciate a second pair of eyes, or another suggested approach to getting this. 

Another potential approach - 

do not render the env provider until the page loads client side - i figured - although it would cause less rerenders, it might cause the initial page loads to slow down, since everything would then be held up from rendering until the page loads client side.

PR #532, opened 2022-08-27
Improve network switching

Schwartz10:
@navFooh one note about this PR - the NetworkSelector story no longer actually switches networks, which is annoying. I tried for ~20 min to get https://www.npmjs.com/package/storybook-addon-next-router working (which would fix network selection), but it wasn't working out of the box, and felt like it wasn't a good use of time to debug that right now. It's working well when i link with the wallet


src/components/NetworkSelector/index.test.tsx
```diff
@@ -101,34 +107,7 @@ describe('NetworkSelector', () => {
...
-        expect(getByText(Network.CALIBRATION)).toBeInTheDocument()
-
-        expect(queryByText(Network.MAINNET)).toBe(null)
-        expect(queryByText(Network.WALLABY)).toBe(null)
-
-        expect(mockRouterPush).toHaveBeenCalledWith(
-          'https://wallet.glif.io/?network=calibration'
-        )
-      })
+      expect(mockRouterPush).toHaveBeenCalledWith('https://wallet.glif.io/')
```
Schwartz10: @navFooh I deleted a good portion of the tests here, which i tried not to, but the reason being that in tests, the network query param and the network end up mismatching after we change networks, which causes the `useEffect` in the `EnvironmentProvider` to fire, switching the network back to the one that matches the URL bar. I tried some hacky workarounds, but i think its better to just check the routerpush call, since the EnvironmentProvider tests the network switching after the fact

src/components/NetworkSelector/index.test.tsx
```diff
@@ -81,12 +92,12 @@ describe('NetworkSelector', () => {
...
     test('it changes the network', async () => {
-      const { getByText, queryByText } = render(
-        <TestEnvironment>
+      const { getByText } = render(
+        <EnvironmentProvider>
           <ThemeProvider theme={theme}>
             <NetworkSelector />
           </ThemeProvider>
-        </TestEnvironment>
+        </EnvironmentProvider>
```
navFooh, line 100: What's the reason for wrapping with `EnvironmentProvider` and not `TestEnvironment`?

Schwartz10: The `TestEnvironment` doesn't (on its own) export `setNetwork` since the EnvironmentProvider does that now

src/utils/potentialDupMsg/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  msg2: Message,
+  currentHeight: number,
+  isPending: boolean = false
+): string | null => {
+  if (!msg1 || !msg2) return null
+
+  const fromAddrEq =
+    isAddrEqual(msg1.from, msg2.from) || isAddrEqual(msg1.from, msg2.from)
+  const toAddrEq =
+    isAddrEqual(msg1.to, msg2.to) || isAddrEqual(msg1.to, msg2.to)
```
navFooh: Doesn't one isAddrEqual check for from and to suffice? They seem identical

src/components/Transaction/Form.tsx
```diff
@@ -27,6 +27,7 @@ import {
   WalletProviderOpts
 } from '../../services'
 import { useLogger } from '../../services/EnvironmentProvider'
+import { WarnPotentialDup } from './PotentialDup'
```
navFooh: Just to be a nitpicker I'd prefer the component and filename to match, either are fine for me

src/components/NetworkSelector/index.test.tsx
navFooh: Ah right. So `Environment` just provides the variables and for stories and tests we don't need more than that I assume. Is that the main reason those two are separated? Would there be something against just ditching `Environment` and only using `EnvironmentProvider`? Just curious.

navFooh (changes requested):

src/components/NetworkSelector/index.test.tsx
Schwartz10: Yeah thats how i've been thinking about it - 

In terms of TestEnvironment using the `EnvironmentProvider` - I think this would be ideal (and something we should work towards), but it's a bigger operation since the current tests mock `useRouter` in different ways, so would need to think about how to consolidate or standardize that mocking. 

Trying to avoid scope creep on this PR so we can get the warn-duplicate stuff in

src/utils/potentialDupMsg/index.ts
Schwartz10: Oh yeah good catch!

PR #533, opened 2022-08-29
Fetch robust address in AddressLink

navFooh:


src/components/LabeledText/AddressLink.tsx
```diff
@@ -18,16 +19,25 @@ export const AddressLink = ({
...
-    if (address) return shouldTruncate ? truncateAddress(address) : address
-    return id || ''
-  }, [address, id, shouldTruncate])
-  const copyText = address || id
+  const { data: gqlAddress } = useAddressQuery({
+    skip: !id || !!address || !fetchAddress,
+    variables: {
+      address: id
+    }
+  })
```
Schwartz10, line 31: Is there use in generalizing here? For instance, instead of _always_ fetching the robust address, should we fetch the _missing_ address type? (robust if id, id if robust)? I'm trying to think of a use case where we'd have the id and not the robust, when a robust actually exists...

Schwartz10 (approved):
See my comment on generalizing...  still merging this so i can unblock you

src/components/LabeledText/AddressLink.tsx
navFooh: When we have the ID and not the robust, is for example when retrieving the list of signers from the MSIG state. There might actually be more cases when lotus returns an ID as address while a robust should be available.

For this component I don't see the use of fetching the ID, because we don't display it when we have the robust.

PR #537, opened 2022-09-01
Message improvements

Schwartz10:
@navFooh one thing in this PR - I put an optional MockApolloProvider in the TestEnvironment so we can start removing any instance of MockApolloProvider in any individual test.


src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -50,26 +42,33 @@ export default function MessageDetail(props: MessageDetailProps) {
...
-    variables: { cid },
-    pollInterval: 10000
-  })
+  const { data: stateReplayQuery, error: stateReplayError } =
+    useStateReplayQuery({
+      variables: { cid: props.cid },
+      pollInterval: 30000,
+      fetchPolicy: 'cache-first',
+      // give the message time to execute on-chain before fetching
+      skip: !props.cid || props.confirmations < 2
```
Schwartz10, line 51: This `!props.cid` check could _probably_ be skipped because the cid prop is required and ts will probably catch this error in any app. But i think it's fine to keep just to limit any mistakes and hit the node with invalid requests to an expensive endpoint

src/components/HistoryTables/MessageHistory/Detail/index.tsx
```diff
@@ -50,26 +42,33 @@ export default function MessageDetail(props: MessageDetailProps) {
...
-    () => gasQuery?.gascost?.gasUsed || 0,
-    [gasQuery?.gascost]
+    () => stateReplayQuery?.stateReplay?.gasCost?.gasUsed || 0,
+    [stateReplayQuery?.stateReplay?.gasCost]
+  )
+
+  const executionTrace = useMemo(
+    () =>
+      (stateReplayQuery?.stateReplay?.executionTrace
+        ?.executionTrace as unknown as ExecutionTrace) || null,
```
Schwartz10, line 70: Since the `executionTrace` comes back as a string and we JSON.parse it before it gets cached, ts isn't happy that i'm converting the types to the parsed ExecutionTrace type. This is ugly, but wasn't sure of the "proper" workaround here (even though i thought we would have the same problem with `message.params`

src/components/HistoryTables/detail.tsx
```diff
@@ -394,11 +399,13 @@ type SeeMoreContentProps = {
...
+  executionTrace: ExecutionTrace
 }
 
 SeeMoreContent.propTypes = {
   message: GRAPHQL_MESSAGE_PROPTYPE.isRequired,
   gasUsed: PropTypes.number.isRequired,
   gasCost: GRAPHQL_GAS_COST_PROPTYPE.isRequired,
-  actorName: PropTypes.string.isRequired
+  actorName: PropTypes.string.isRequired,
+  executionTrace: EXECUTION_TRACE_PROPTYPE.isRequired
```
Schwartz10, line 410: You'll notice i added this and pass the prop for now, but it's not being used yet. I'll follow up in another PR with the additional UIs to show subcals

navFooh (approved):

PR #538, opened 2022-09-01
Explorer update

navFooh:


src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -39,111 +35,96 @@ export default function ProposalDetail({
...
 
+  // Load the actor state
   const {
     data: actorData,
     loading: actorLoading,
+    notFound: actorNotFound,
     error: actorError
-  } = useActorQuery({
-    variables: { address: convertAddrToPrefix(msigAddress, coinType) }
-  })
```
navFooh, line 70: I'm not sure why to `useActorQuery` was used here. It seemed to only grab the actor code, to confirm that it was an msig actor and then perfom `useStateReadState`, which also gets the actor code

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -39,111 +35,96 @@ export default function ProposalDetail({
...
+    return isMsigActor && actorData ? (actorData.State as MsigState) : null
+  }, [actorData, actorName])
 
-  const {
-    data: stateData,
-    loading: stateLoading,
-    error: stateError
-  } = useStateReadStateQuery<MsigState>({
-    variables: { address: convertAddrToPrefix(msigAddress, coinType) },
-    skip: !loadMsigState
```
navFooh, line 90: Skip was not being handled by `useStateReadStateQuery`, see comment in `useStateReadState.ts`

src/components/LabeledText/AddressLink.tsx
```diff
@@ -10,45 +10,43 @@ import {
...
+  ...labeledLinkProps
 }: AddressLinkProps) => {
   const { explorerUrl } = useEnvironment()
   const { data: gqlAddress } = useAddressQuery({
-    skip: !id || !!address || !fetchAddress,
-    variables: {
-      address: id
-    }
+    skip: !fetchAddress,
+    variables: { address }
```
navFooh, line 32: Changed the logic here a bit, guessing this is what you mentioned before :)

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,204 @@
...
+      if (Name === 'Cid') {
+        const cid = Children['/'].Value as string
+        return (
+          <DataTypeLine
+            label={`${label} (CID)`}
+            depth={depth}
+            dataType={{ Name, Type: Type.String, Value: cid }}
+          />
+        )
+      }
```
navFooh, line 58: This is temporary until we really know how to unwrap the CIDs

src/utils/useMsigGetAvailableBalance/index.ts
```diff
@@ -0,0 +1,38 @@
...
+        .then((b: string) => setBalance(new FilecoinNumber(b, 'attofil')))
+        .catch((e: Error) => setError(e))
+        .finally(() => setLoading(false))
+    } else {
+      setError(new Error('Invalid address'))
+    }
+  }, [address, lotusApi])
+
+  return { availableBalance: balance, loading, error }
+}
```
navFooh: Separated this from useStateReadState

src/utils/useStateReadState/index.ts
```diff
@@ -1,123 +1,45 @@
...
-}
-
-/////// implemented until https://github.com/glifio/graph/issues/33
-export const useStateReadStateQuery = <T = any>(
-  baseOptions: QueryHookOptions<
-    T,
-    {
-      address: string
-    }
-  >
```
navFooh, line 38: The name and params made it seem like this was an apollo query, which it's not. The params like `skip` that could be passed were not handled either, only the address was grabbed from the `baseOptions`.

src/utils/useStateReadState/index.ts
```diff
@@ -1,123 +1,45 @@
...
-        } else setError(new Error('Failed to fetch'))
-      } finally {
-        setLoading(false)
-      }
-    }
-
-    const firstLoad = !actorState && loading && !error
-    const newLoad = fetchedFor !== baseOptions?.variables?.address && !error
-
-    if (firstLoad || newLoad) {
```
navFooh: This logic is removed now, but I wonder if we really need it. I can reimplement something similar if you think we do

src/utils/useStateReadState/index.ts
```diff
@@ -1,123 +1,45 @@
...
-    fetchedFor,
-    setFetchedFor,
-    apolloClient,
-    lotusApi,
-    networkName
-  ])
+  }, [address, lotusApi])
 
-  return { data: actorState, error, loading }
+  return { data, loading, notFound, error }
```
navFooh: Return a `notFound` property instead of an empty object. Then we can show in the frontend what is going on instead of displaying an actor page with empty values. I was confused about this before.

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,118 +1,166 @@
...
   const {
     data: addressData,
-    error: addressError,
-    loading: addressLoading
+    loading: addressLoading,
+    error: addressError
   } = useAddressQuery({
-    variables: { address: convertAddrToPrefix(address, coinType) }
+    variables: { address }
   })
```
Schwartz10, line 54: Not sure if we want to slow this down to prevent errors, but just to point out...

In some cases, a user might be viewing an SECP256K1 (`f1`) or BLS (`f3`) actor that does not yet exist on chain (no transactions have been sent from / to it). In this case, `StateReadState` would error 

However, thinking about it more, i think we should leave as-is - probably better to parallelize this and you've already handled the actor not found error

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,118 +1,166 @@
...
 
-  const [viewActorState, setViewActorState] = useState(false)
+  // Get the actor name from the actor code
+  const actorName = useMemo<string | null>(
```
Schwartz10, line 57: Do you think it would be useful to abstract the bulk of this logic out into its own `useActor` hook which could take an `address` and return the complete decorated actor state if it exists? It could also do some smarter (and aggregated) caching and polling. 

See my comment in the general review about redundant data fetching - if we stuck a `useActor` query into the react-components, the safe could just `useActor` too and share in the caching and revalidation benefits

The `proposal.tsx` would also benefit

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,118 +1,166 @@
...
+          )}
           {addressData?.address.robust && (
             <Line label='Robust address'>{addressData?.address.robust}</Line>
           )}
-          {addressData?.address.id && (
-            <Line label='ID'>{addressData?.address.id}</Line>
-          )}
-          <Line label='Actor'>{actorType}</Line>
+          <Line label='Actor name'>{actorName || 'unknown'}</Line>
+          <Line label='Actor code'>{actorData.Code['/']}</Line>
```
Schwartz10, line 139: Can we `text-transform: capitalize` this?

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,118 +1,166 @@
...
         error={error}
       />
-      {!loading && !error && (
+      {!loading && !error && !actorNotFound && (
         <Lines>
+          {addressData?.address.id && (
+            <Line label='ID address'>{addressData?.address.id}</Line>
+          )}
           {addressData?.address.robust && (
             <Line label='Robust address'>{addressData?.address.robust}</Line>
```
Schwartz10: Let's reverse these (show robust first)

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,118 +1,166 @@
...
           )}
-          {addressData?.address.id && (
-            <Line label='ID'>{addressData?.address.id}</Line>
-          )}
-          <Line label='Actor'>{actorType}</Line>
+          <Line label='Actor name'>{actorName || 'unknown'}</Line>
+          <Line label='Actor code'>{actorData.Code['/']}</Line>
           <Line label='Balance'>
-            {new FilecoinNumber(actorStateData?.Balance, 'attofil').toFil()} FIL
+            {new FilecoinNumber(actorData.Balance, 'attofil').toFil()} FIL
```
Schwartz10: Let's use `makePrettyBalance` here, 6 decimals

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -1,118 +1,166 @@
...
-                (actorStateData?.State as MsigState).AvailableBalance,
-                'attofil'
-              ).toFil()}{' '}
-              FIL
+              {availableBalanceError ? (
+                <>Failed to load</>
+              ) : availableBalanceLoading ? (
+                <>Loading...</>
+              ) : availableBalance ? (
+                <>{availableBalance.toFil()} FIL</>
```
Schwartz10, line 148: Let's use `makePrettyBalance` here, 6 decimals

src/components/HistoryTables/MsigProposals/Proposal.tsx
Schwartz10: Probably because i'm dumb, but possibly because I think originally i envisioned this working in a way where a user could enter any actor address (i.e. showing the proposals on the explorer) - making the `useActorQuery` helped me get a better error about "actor not found" at the time. Good call here though

src/components/LabeledText/AddressLink.tsx
```diff
@@ -10,45 +10,43 @@ import {
   LabeledLinkPropTypes
 } from './LabeledLink'
 
+/**
+ * AddressLink is a LabeledLink for actor addresses.
+ *
+ * When fetchAddress is false (default), both id and address can be
```
Schwartz10, line 16: 2 Small nits:

1. Think the comment line here should be changed from: "but only the (robust) address will be shown if present" to "but the robust address takes precedent over the id address if they both exist"
2. i know this wasn't introduced in this PR, and it's probably inconsistent in other places too, but I feel like we should move towards naming consistency in:

```
address: {id: string, robust: string}
robust: string
id: string
```

The AddressLink uses `address` as a string, which can either be a robust or an ID, which feels inconsistent. 

Maybe we can add an optional `robust` param, and make the `address` type _either_ the GQL address type or a string, and handle both, so we can start using it correctly.

Happy to open a separate ticket for this thought if you agree

src/components/LabeledText/AddressLink.tsx
Schwartz10: Yes

src/components/LabeledText/MessageLink.tsx
```diff
@@ -9,16 +9,11 @@ import {
   LabeledLinkPropTypes
 } from './LabeledLink'
 
-export const CidLink = ({
+export const MessageLink = ({
```
Schwartz10, line 12: Good call on this, was having similar thoughts before i saw this!

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,283 @@
...
+  depth,
+  dataTypeMap
+}: DataTypeMapLinesProps) => (
+  <>
+    {label && <Line label={label} depth={depth} />}
+    {Object.entries(dataTypeMap).map(([key, dt]) => (
+      <DataTypeLines
+        key={key}
+        label={key}
+        depth={label ? (depth ?? 0) + 1 : depth}
```
Schwartz10, line 97: Think we could simplify things a bit if we give `depth` a `defaultProps` value of `1` - since (i think?), the only time you'd need to include a >1 `depth` value is recursively... the developer him/herself probably doesn't need to pass depth unless in a special case that we don't yet have?

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,283 @@
...
+ * Renders the value of a DataType using the most suitable component.
+ */
+
+export const DataTypeValue = ({ dataType }: DataTypeValueProps) => {
+  const { Name, Value } = dataType
+
+  switch (dataType.Type) {
+    case Type.Bool:
+      const boolVal = Value as boolean
+      return <>{boolVal.toString()}</>
```
Schwartz10, line 150: lol news to me... the boolean primitive has a toString method

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,283 @@
...
+    case Type.String:
+      const strVal = Value as string
+
+      if (Name === 'Address')
+        return (
+          <AddressLink address={strVal} fetchAddress hideCopyText={false} />
+        )
+
+      if (Name === 'FilecoinNumber')
+        return <>{new FilecoinNumber(strVal, 'attofil').toFil() + ' FIL'}</>
```
Schwartz10: Let's `makeFriendlyBalance` with 6 decimals here?

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+      return <>{Value}</>
+
+    case Type.Number:
```
Schwartz10, line 167: I think all Number types in the VM use 18 decimal precision, so they'll _all_ be in the `atto` SI prefix unit, meaning we should still be able to 

```
return <>{new FilecoinNumber(strVal, 'attofil').toFil()}
```

And just leave it without the "FIL" denominator at the end like the FilecoinNumber type has

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,283 @@
...
+      const numVal = Value as number
+      return <>{numVal}</>
+
+    case Type.Array:
+      const arrVal = Value as Array<BaseValue>
```
Schwartz10: In our current use, can this ever actually happen? It looks like `DataTypeLines` maps over arrays and renders a `DataTypeLine` for each 

Either way, it seems like we should share similar behavior, and instead of joining array values by a string, we should just recurse through the array and render `DataTypeLines` for each array item like we do elsewhere?

Hard for me to get a good sense since i'm having trouble finding a case where this hits

src/components/Layout/DataTypes.tsx
```diff
@@ -0,0 +1,283 @@
...
+      )
+
+    case 'bigint':
+    case 'boolean':
+    case 'function':
+    case 'number':
+    case 'symbol':
+      return (
+        <Line label={label} depth={depth}>
+          {data.toString()}
```
Schwartz10, line 205: What happens if we `toString` a function?

src/utils/useStateReadState/index.ts
```diff
@@ -1,123 +1,45 @@
...
-  const { lotusApi, networkName } = useEnvironment()
-  const [actorState, setActorState] = useState<LotusRPCActorState<T> | null>(
-    null
-  )
-  const [loading, setLoading] = useState(true)
-  const [error, setError] = useState<Error>(undefined)
-  const [fetchedFor, setFetchedFor] = useState<string>('')
-
-  const apolloClient = useApolloClient()
+export const useStateReadState = <T = object | null>(
```
Schwartz10, line 15: I think we should refactor to a `useSWR` approach here, like we should for `useMsigAvailablebalance` - comes equipped with more tools like polling, caching, and revalidation that will be useful for keeping actor states dynamically updating

Schwartz10 (changes requested):
1. The Proposal UI and the ActorState UI both use the "see more" dropdown links, but they look inconsistent:

image

image


Should we just take out the `<hr />`s that wrap the See More in the proposal detail?

2. Maybe better for a follow-on PR, but it seems like we do a lot of redundant data fetching in `Proposal.tsx` - this page currently _only_ exists inside the Safe, where the ActorName, ActorState...etc already exist. Seems like this could be solved nicely with a `useActor` hook that does a lot of the redundant work and caches + revalidates it
3. I think we should do a brief refactor of `useMsigGetAvailableBalance` and `useStateReadState` to a `useSWR` based approach - generally i've found it quite reliable, especially with built-in polling and revalidation functionality (which seems important for a dynamically refreshing /home UI in the safe). See example here https://github.com/glifio/react-components/blob/primary/src/services/UpdateBalance/index.tsx, makes it easy to write tests for this too
4. Testing - i think we should write some brief and simple tests for `useMsigAvailableBalance`, `useStateReadState`, and `ActorState` - hopefully the hooks testing should be quite simple because we have an example (thats probably more complicated) in `BalancePoller`, and we can mock most of it quite easily

navFooh:
> 1. The Proposal UI and the ActorState UI both use the "see more" dropdown links, but they look inconsistent:
My proposal would be to remove the "see more" toggle from the proposal altogether and just display the information that we're hiding by default. My understanding was that we want to limit the amount of space that the actor state consumes vertically so that we can still see the tx history below, but I don't feel this goes for the proposal. What it's hiding now is the next transaction ID and the list of people who already approved, which doesn't seem like too much info to me.

> 2. Maybe better for a follow-on PR, but it seems like we do a lot of redundant data fetching in `Proposal.tsx`
Let's do a follow up PR, I've created an issue: https://github.com/glifio/react-components/issues/548

> 3. I think we should do a brief refactor of `useMsigGetAvailableBalance` and `useStateReadState` to a `useSWR` based approach
Will have a look at this

> 4. Testing - i think we should write some brief and simple tests for `useMsigAvailableBalance`, `useStateReadState`, and `ActorState`
Will have a look at this

src/components/Layout/DataTypes.tsx
navFooh: It should probably never occur that we have to display javascript functions here, but since this is like a fallback that should work in all cases, it doesn't hurt either.

```javascript
function test (foo, bar) {
    return foo + bar
}
test.toString() // function test (foo, bar) {\n    return foo + bar\n}
```

src/components/Layout/DataTypes.tsx
navFooh: That's true. I was considering a use case where we do want to display an array on a single line (maybe a bitfield), but that's not really relevant right now. I'll remove it and we can tackle it again if we need it

src/components/Layout/DataTypes.tsx
navFooh: I'm not sure if I understand correctly. When `dataType.Type === Type.Number`, it should just be a regular number, like a transaction ID or something, but not related to Filecoin values. Then it should always be a string.

src/components/HistoryTables/ActorState/index.tsx
navFooh: Created an issue for this: https://github.com/glifio/react-components/issues/548

src/components/HistoryTables/ActorState/index.tsx
navFooh: Yeah I imagine it will just show the actor not found. We can follow up with another PR if there is a better way to tackly this

src/components/Layout/DataTypes.tsx
Schwartz10: Ahh ok got it. I thought you were using Number for anything that wasn't specifically a Filecoin balance

Schwartz10 (changes requested):
Hey this looks good @navFooh - two things:

1. Let's use an `useSWR` approach for `useMsigAvailableBalance` and `useStateReadState`
2. Let's add (just 1 is fine) test for `useMsigAvailableBalance` and `useStateReadState` (data should be mocked)
3. Let's add (probably 1 is fine as well) test for `ActorState`

Once you get those in, we can squash, merge, republish

PR #542, opened 2022-09-05
Fix 541

### LINKED ISSUE #541: Pending message should remove badge
image

Schwartz10:
Close #541

### DIFF
--- a/src/components/HistoryTables/MessageHistory/History/MessagePendingRow.tsx
+++ b/src/components/HistoryTables/MessageHistory/History/MessagePendingRow.tsx
@@ -34,11 +34,7 @@ export default function PendingMessageHistoryRow(
           {truncateAddress(message.cid)}
         </SmartLink>
       </td>
-      {props.inspectingAddress && (
-        <td>
-          <Badge color='purple' text={methodName} />
-        </td>
-      )}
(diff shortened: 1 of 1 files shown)


src/components/HistoryTables/MessageHistory/History/MessagePendingRow.tsx
```diff
@@ -34,11 +34,7 @@ export default function PendingMessageHistoryRow(
           {truncateAddress(message.cid)}
         </SmartLink>
       </td>
-      {props.inspectingAddress && (
-        <td>
-          <Badge color='purple' text={methodName} />
-        </td>
-      )}
+      {props.inspectingAddress && <td className='capitalize'>{methodName}</td>}
```
navFooh: I don't see any styling element that would use the `capitalize` class to capitalize the method name? I see that in the `MessageConfirmedRow` you added a `MessageHistoryRowWrapper` intead of the `<tr>` to style the `<td>` inside.

Tbh I'd just do something like this:
```
const CapitalizedTD = styled.td`
  text-transform: capitalize;
`
```

And use that instead of `<td className='capitalize'>`. It seems a bit unnecessary to style the TR to select a TD classname when the TR or any other component inside don't need any special styling.

navFooh (changes requested):

PR #549, opened 2022-09-07
Make friendly balance update

navFooh:


src/components/BalanceCard/index.tsx
```diff
@@ -6,7 +6,7 @@ import { FilecoinNumber } from '@glif/filecoin-number'
 import { ButtonV2 } from '../Button/V2'
 import { OutlineBox } from '../Layout'
 import { FILECOIN_NUMBER_PROPTYPE } from '../../customPropTypes'
-import makeFriendlyBalance from '../../utils/makeFriendlyBalance'
+import { makeFriendlyBalance } from '../../utils'
```
Schwartz10: I think we should leave the more specific import here, this way we don't _necessarily_ bring in the entire `utils` directory when we import `BalanceCard`

src/components/Transaction/Balance.tsx
```diff
@@ -4,8 +4,7 @@ import {
   ADDRESS_PROPTYPE,
   FILECOIN_NUMBER_PROPTYPE
 } from '../../customPropTypes'
-import truncateAddress from '../../utils/truncateAddress'
-import makeFriendlyBalance from '../../utils/makeFriendlyBalance'
+import { makeFriendlyBalance, truncateAddress } from '../../utils'
```
Schwartz10: Same thing here, I'd leave the direct imports

src/utils/makeFriendlyBalance/index.test.js
```diff
@@ -1,123 +1,301 @@
-import { BigNumber } from '@glif/filecoin-number'
-import makeFriendlyBalance from '.'
+import { FilecoinNumber } from '@glif/filecoin-number'
+import { makeFriendlyBalance } from '.'
```
Schwartz10, line 2: And here

src/utils/makeFriendlyBalance/index.ts
```diff
@@ -1,65 +1,61 @@
...
+  let power = 0
+  const units = ['K', 'M', 'B']
+  for (const unit of units) {
+    power++
+    const unitVal = dpValue.dividedBy(Math.pow(1000, power))
+    if (unitVal.isLessThan(1000))
+      return (
+        minus + unitVal.decimalPlaces(3, BigNumber.ROUND_DOWN).toString() + unit
+      )
   }
```
Schwartz10, line 58: So much nicer!

Schwartz10 (changes requested):
Looks good - think we just fix those imports and then merge

PR #562, opened 2022-09-14
Describe params

navFooh:
Note that I've updated all the useSWR hooks to include `lotusAPI` and `lotusMethod` in the key. Previously the fetcher was sometimes updated with `useCallback` when the `lotusApi` changed from environment, but changing the fetcher does not update cached data in SWR. If we don't include `lotusAPI` as a key, we'll probably get weird data issues when switching networks, `lotusMethod` is good to have for sanity as some methods would probably take the same argument (like an actor address).


.storybook/preview.jsx
```diff
@@ -13,7 +14,9 @@ export const decorators = [
     <TestEnvironment>
       <ApolloWrapper>
         <ThemeProvider theme={theme}>
-          <Story />
+          <ErrorBoundary>
+            <Story />
+          </ErrorBoundary>
```
navFooh, line 19: I find it useful to wrap the story in an ErrorBoundary. This way you can still open a story that errors and see the error in the console. Otherwise storybook will omit the story from the menu altogether and pretend it doesn't exist.

package.json
```diff
@@ -17,9 +17,7 @@
     "test:jest": "jest --coverage",
     "test-watch": "jest --watch",
     "clean": "rimraf dist/* && mkdirp dist",
-    "codegen": "run-s codegen:*",
-    "codegen:actorCids": "node src/scripts/generate_actor_codes.mjs && eslint ./src/generated/actorCodes.ts --fix",
-    "codegen:graphql": "graphql-codegen --config codegen.yml && eslint ./src/generated/graphql.tsx --fix",
+    "codegen": "graphql-codegen --config codegen.yml && eslint ./src/generated/graphql.tsx --fix",
```
navFooh, line 20: These are now coming from `filecoin-actor-utils`

src/components/ErrorBoundary/index.tsx
```diff
@@ -84,7 +84,7 @@ const propTypes = {
 }
 
 const defaultProps = {
-  Wrapper: ({ children }) => <>{children}</>,
+  Wrapper: <></>,
```
navFooh: I found that a default value of a functional component instead of plain JSX gave an error in Storybook

src/components/HistoryTables/MessageHistory/History/MessageConfirmedRow.tsx
```diff
@@ -33,17 +25,17 @@ export default function MessageHistoryRow(props: MessageHistoryRowProps) {
...
   return (
-    <MessageHistoryRowWrapper>
+    <tr>
       <td>
         <SmartLink href={cidHref(message.cid)}>
           {truncateAddress(message.cid)}
         </SmartLink>
       </td>
-      <td className='capitalize'>{methodName}</td>
+      <td>{methodName}</td>
```
navFooh, line 37: methodName is now returned as e.g. `AddSigner`. We can add a regex to split these words so it says `Add Signer`, but I'd propose to do that in `filecoin-actor-utils`

src/components/NetworkSelector/useNetworkName.ts
```diff
@@ -1,22 +1,24 @@
...
+
+export const useNetworkName = (): UseNetworkNameResult => {
+  const { lotusApi } = useEnvironment()
+  const { data, error } = useSWR<string, Error>(
+    [lotusApi, 'StateNetworkName'],
+    fetcher
+  )
 
-  return { networkName, error, loading: isValidating }
+  return { networkName: data ?? '', loading: !data && !error, error }
```
navFooh, line 27: `isValidating` is also true when there is a `refreshInterval`. According to the docs it's best to check for the initial loading state with `!data && !error`

src/services/EnvironmentProvider/index.tsx
```diff
@@ -18,8 +18,8 @@ import { appendQueryParams, getQueryParam, removeQueryParam } from '../../utils'
 
 export enum Network {
   MAINNET = 'mainnet',
-  CALIBRATION = 'calibration',
-  WALLABY = 'wallaby'
+  CALIBRATION = 'calibrationnet',
+  WALLABY = 'wallabynet'
```
navFooh, line 22: I've changed these to the offical names, it makes a bit more sense to me. Like it's not `main`, but `mainnet`, so also not `calibration` but `calibrationnet`.

It also prevents us from having to fetch the official name when switching network and I don't think we really need to use `useNetworkName` in the `NetworkSwitcher` anymore, but wanted to check what you thought about this first.

For me it's fine to keep using the `useNetworkName` hook, but then I'd say we expose the official network name from the `EnvironmentProvider` as well.

src/components/ErrorBoundary/index.tsx
navFooh: Nevermind, using the plain JSX tag gives an error with the testing library

src/components/HistoryTables/ActorState/index.tsx
```diff
@@ -69,12 +69,16 @@ export const ActorState = ({ address: addressProp }: ActorStateProps) => {
...
   const describedState = useMemo<DataTypeMap | null>(() => {
     try {
-      return actorData ? describeLotusActorState(actorData) : null
+      return actorName && actorData?.State
+        ? describeActorState(actorName, actorData.State)
+        : null
     } catch (e) {
-      logger.error(e)
+      logger.error(
+        `Failed to describe actor state for address ${address}: ${e.message}`
```
Schwartz10: Might be helpful to have the `networkName` here

src/components/HistoryTables/MessageHistory/History/MessageConfirmedRow.tsx
Schwartz10: Low priority

src/components/HistoryTables/MessageHistory/hooks/useMethodName.ts
```diff
@@ -1,46 +1,25 @@
...
+  const { data } = useActorQuery({
     variables: {
-      address: convertAddrToPrefix(
-        message?.to.robust || message?.to.id,
-        coinType
-      )
-    }
+      address: convertAddrToPrefix(address?.robust || address?.id, coinType)
+    },
+    skip: !address
```
Schwartz10, line 18: Technically, if no `id` address exists, the `useActorQuery` will return null. I think this is fine for now, but we _could_ be more specific about when to skip in the future. An `id` gets assigned when a `robust` (stable) address has some interaction with the chain

Schwartz10: Additional future optimization - we know that an `f1` or an `f3` actor _must_ be an `account` actor. An `f2` address is a non-account actor

src/components/HistoryTables/MessageHistory/hooks/useMethodName.ts
```diff
@@ -1,46 +1,25 @@
...
-  const methodName = useMemo(() => {
-    if (actorName) return getMethodName(actorName, Number(message.method))
-    else return '...'
-  }, [actorName, message?.method])
-
-  return { methodName, actorName }
+  // Resolve actor code, name and message name
+  const actorCode = data?.actor?.Code
+  const actorName = actorCode ? getActorName(actorCode, networkName) : null
+  return actorName ? getMethodName(actorName, Number(method)) : null
```
Schwartz10, line 24: Think this is fine, but seems like this should use `code` here in the future, since most methods won't be associated with a uniquely named built-in actor. Can tackle this in a future sprint

src/components/HistoryTables/MsigProposals/Proposal.tsx
```diff
@@ -161,40 +168,30 @@ export default function ProposalDetail({
...
-              value: proposal.value,
-              method: proposal.method,
-              params: proposal.params
-            }
-          }}
-          actorName='/multisig'
-          depth={0}
+        <hr />
+        <AddressLine label='To' value={proposal.to} />
+        <MethodLine
```
Schwartz10, line 182: I think we can just use a regular `Line` for this. The badges look kinda goofy i think

src/components/HistoryTables/MsigProposals/ProposalRow.tsx
```diff
@@ -52,7 +53,7 @@ export default function ProposalHistoryRow({
       <td>
         <Badge
```
Schwartz10: Here I think we can do away with the badge too

src/components/Layout/Lines.tsx
```diff
@@ -146,6 +153,55 @@ NullishLine.propTypes = {
...
+ */
+
+export const MethodLine = ({
+  label,
+  depth,
+  actorName,
+  methodNum
+}: MethodLineProps) => (
+  <Line label={label} depth={depth}>
+    <Badge
```
Schwartz10: Think the Badge is unnecessary here as mentioned above - component still might be nice to have

src/components/Layout/LinesParams.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }, [actorCode, networkName, logger])
+
+  // Get described parameters
+  const describedParams = useMemo<DataType | null>(() => {
+    if (actorName && decodedParams) {
+      try {
+        return describeMessageParams(actorName, method, decodedParams)
+      } catch (e) {
+        logger.error(
+          `Failed to describe message params for actor address: ${address}, method: ${method}, params: ${params}, with message: ${e.message}`
```
Schwartz10: networkName might be helpful here

src/components/Layout/LinesParams.tsx
```diff
@@ -0,0 +1,89 @@
...
+        )
+      }
+    }
+    return null
+  }, [address, method, params, actorName, decodedParams, logger])
+
+  // Return most verbose params first
+  return describedParams ? (
```
Schwartz10, line 66: Nicely done 👏

src/components/NetworkSelector/useNetworkName.ts
```diff
@@ -1,22 +1,24 @@
...
-  } = useSWR(lotusApiAddr, fetcher)
+const fetcher = async (
+  lotusApi: LotusRpcEngine,
+  lotusMethod: string
+): Promise<string> => await lotusApi.request<string>(lotusMethod)
+
+export const useNetworkName = (): UseNetworkNameResult => {
+  const { lotusApi } = useEnvironment()
+  const { data, error } = useSWR<string, Error>(
+    [lotusApi, 'StateNetworkName'],
```
Schwartz10: I'm not sure about passing an object as the key here... without reading the docs, gut instincts tell me to use the `lotusApiAddr` string here and instantiate the lotusApi inside the `fetcher`

I know they use those keys for caching

src/services/EnvironmentProvider/index.tsx
Schwartz10: I agree we should use the official names, but the only issue is now some documentation points to the `?network=wallaby` or `?network=calibration`, so we should just handle those cases appropriately to not break backwards compatibility with those query params

src/services/UpdateBalance/index.tsx
```diff
@@ -1,47 +1,56 @@
...
-    },
-    [updateBalance, lotusApi, logger]
-  )
-
-  const { data } = useSWR(
-    wallet.address ? [wallet.address, wallet.balance, selectedWalletIdx] : null,
+  const { lotusApi } = useEnvironment()
+  const { selectedWalletIdx, updateBalance } = useWalletProvider()
+  const { data, error } = useSWR<FilecoinNumber, Error>(
+    [lotusApi, 'WalletBalance', wallet.address],
```
Schwartz10: same thing about my comment above - better to use a string here?

src/utils/useStateDecodeParams/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // Temporarily attempt to parse as JSON string,
+  // until StateDecodeParams is removed from graph
+  try {
+    return JSON.parse(base64Params)
+  } catch {
+    // params are not a JSON string
+  }
+
+  return await lotusApi.request<T>(
```
Schwartz10, line 34: Eventually i think we should move this call to the graph and cache it since i imagine we'll be making the same call over and over

src/utils/useStateDecodeParams/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  )
+}
+
+export const useStateDecodeParams = <T = Record<string, any> | null>(
+  address: string,
+  method: number,
+  params: string
+): UseStateDecodeParamsResult<T> => {
+  const { lotusApi } = useEnvironment()
+  const { data, error } = useSWR<T, Error>(
```
Schwartz10: here it probably makes sense to use https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations

src/components/HistoryTables/detail.tsx
```diff
@@ -10,8 +10,7 @@ import {
   useChainHeadSubscription
 } from '../../generated/graphql'
 import { IconCheck, IconPending, IconClock } from '../Icons'
-import { Badge, Lines, Line, AddressLine } from '../Layout'
-import { getMethodName } from './methodName'
+import { Badge, Line } from '../Layout'
```
Schwartz10, line 13: Here we might as well remove the Badge too...

src/utils/useStateReadState/index.ts
```diff
@@ -12,33 +12,30 @@ interface UseStateReadStateResult<T> {
...
-        actorAddress,
-        null
-      )
-    },
-    [lotusApi]
-  )
-
   const { data, error } = useSWR<LotusActorState<T>, Error>(
-    [address, 'StateReadState'],
+    [lotusApi, 'StateReadState', address],
```
Schwartz10: marking here for string usage

Schwartz10 (changes requested):
Hey @navFooh this looks good. Two general comments:

1. I left comments about getting rid of the badge in this PR review. If you'd rather we tackle this in a follow on PR thats fine by me
2. I see you can pass objects as keys in useSWR https://swr.vercel.app/docs/arguments#passing-objects, but it feels safer to me to just pass the string and instantiate the LotusAPI inside the `fetcher` callbacks. I dont feel strongly about this if you disagree, thats fine

src/services/EnvironmentProvider/index.tsx
navFooh: Good call, I've updated it so it accepts with or without the "net" suffix

src/components/HistoryTables/ActorState/index.tsx
navFooh: Good call

src/utils/useStateDecodeParams/index.ts
navFooh: Yeah I think we should move it to the graph as well but then we don't need to call 'StateDecodeParams' from Lotus, we can use the GoLang types to unmarshall CBOR and marshall to JSON. We should do this when we circle back to this scope of work.

src/components/NetworkSelector/useNetworkName.ts
navFooh: Yeah I think you're right it's better to use the string. In theory it's fine to use the lotusApi as a key, but since the `lotusApi` is reinstantiated between network switches, it won't keep the cache. When using the api URL as string, the cache would still be valid after switching back to a network.

src/utils/useStateDecodeParams/index.ts
navFooh: looks good!

src/components/NetworkSelector/useNetworkName.ts
```diff
@@ -9,14 +9,18 @@ interface UseNetworkNameResult {
...
 const fetcher = async (
-  lotusApi: LotusRpcEngine,
+  apiAddress: string,
   lotusMethod: string
-): Promise<string> => await lotusApi.request<string>(lotusMethod)
+): Promise<string> => {
+  if (!apiAddress) return null
+  const lotusApi = new LotusRpcEngine({ apiAddress })
+  return await lotusApi.request<string>(lotusMethod)
+}
```
Schwartz10: Mental note - a nice lil generic function now

src/components/NetworkSelector/useNetworkName.ts
```diff
@@ -19,7 +19,7 @@ const fetcher = async (
 
 export const useNetworkName = (): UseNetworkNameResult => {
   const { lotusApiUrl } = useEnvironment()
-  const { data, error } = useSWR<string, Error>(
+  const { data, error } = useSWRImmutable<string, Error>(
```
Schwartz10: Here i think we do actually want to use `useSWR` because we want to know if we disconnect from the node 

i'll fix

Schwartz10 (approved):

PR #563, opened 2022-09-16
Return values

navFooh:


src/components/Layout/LinesParams.tsx
```diff
@@ -68,8 +68,12 @@ export const LinesParams = ({ address, method, params }: LinesParamsProps) => {
   ) : decodedParams ? (
     <BaseTypeObjLines label='Params' data={decodedParams} />
   ) : params ? (
-    <Line label='Params'>
-      {params} ({decodeParamsError ? 'failed to decode' : 'decoding...'})
+    <Line
+      label={`Params (${
+        decodeParamsError ? 'failed to decode' : 'decoding...'
+      })`}
+    >
```
navFooh, line 75: Looks a bit nicer to have the info next to the label instead of the base64 data I think

src/components/Layout/LinesReturn.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // Get actor code and name
+  const actorCode = useMemo(() => actorData?.actor?.Code, [actorData])
+  const actorName = useMemo<string | null>(() => {
+    if (actorCode) {
+      const name = getActorName(actorCode, networkName)
+      if (name) return name
+      logger.error(`Failed to retrieve actor name for code: ${actorCode}`)
+    }
+    return null
+  }, [actorCode, networkName, logger])
```
navFooh, line 51: This is maybe a bit repetitive as we do the same in `LinesParams.tsx`, but I think we should leave it for now until we revise circle back to the explorer scope

src/components/Layout/LinesReturn.tsx
```diff
@@ -0,0 +1,95 @@
...
+    decodedReturnVal,
+    networkName,
+    logger
+  ])
+
+  // Return most verbose return value first
+  return describedReturnVal ? (
+    <DataTypeLines label='Return' dataType={describedReturnVal} />
+  ) : returnVal ? (
+    <Line label='Return (failed to decode)'>{returnVal}</Line>
```
navFooh, line 79: Here I don't think we should show the `decodedReturnVal`, since it's mostly Uint8Arrays that don't display well anyway. If describedReturnVal fails, we can consider the decoding as failed as well, since some decoding happens in there now

Schwartz10: Yeah that makes sense to me

PR #576, opened 2022-09-29
Cleanup v1

navFooh:
This PR removes all unused and deprecated definitions from `theme.js`. This also means that there is no more reason to wrap everything in the `ThemeProvider`. If we ever want to do things like dark mode, we can do so by changing the CSS variables depending on a body class for example.


src/services/WalletProvider/metamaskUtils/state.ts
```diff
@@ -1,4 +1,4 @@
-import { bool } from 'prop-types'
+import PropTypes from 'prop-types'
```
Schwartz10, line 1: I'm fine leaving this, but why do you prefer this way? To keep things consistent?

My understanding is that by using destructing here ,we can avoid pulling in objects we don't need rather than importing everything and only access the `bool` property on the export default of the prop-types package. Although, I'm sure it's not that simple under the hood?

src/components/theme.ts
```diff
@@ -1,395 +1,13 @@
-export const baseColors = {
```
Schwartz10, line 1: 🥰🥰

src/components/CopyText/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  .copy-text {
+    position: absolute;
+    top: calc(50% - 0.2em);
+    left: calc(100% + 0.25em);
+    opacity: 0;
+    color: ${props => props.color};
+    transform: translateY(-50%);
+    transition: opacity 0.2s ease-in;
+    pointer-events: none;
```
Schwartz10, line 25: I _think_ this might be changing the behavior of the old CopyText component, and I personally prefer the old behavior.

The old behavior - hovering over the CopyIcon _or_ the <span> tag would cause the "copy"/"copied" text to stay visible.

The new behavior - hover over the CopyIcon causes the "copy"/"copied" text to stay visible. When i scroll right to click "copy", it disappears, which is kinda annoying.

Removing `pointer-events: none` seems to restore the old behavior, but i held off on making this change because I wasn't sure if keeping this css property was important for some other reason? Or if there was a better approach to restoring the old behavior

Schwartz10 (commented):
Hey @navFooh great job on getting this over the finish line! 

I added a commit which standardizes the typescript `enum` as a wrapper around our base-css variables. 

Also changes the tooltip pointer to be the help icon rather than a pointer (which makes it feel clickable, when it's not)

**I left you the honors of `npm uninstall styled-system` :D** - let's get that out of the package.json and package-lock 

In the future, I think we could get rid of the third "walkthrough" step for creating seed phrases (the one where it says "success! click next to go to the wallet" - seems useless. I tried (very quickly) to rip that step out, but it was a bit harder to get the tests to resume so i kept it for now since i believe a very small % of users actually create seed phrases in the wallet

if you're good with my changes, once we restore the old "Copy" button behavior, feel free to squash and merge this. I didn't link this to the safe and wallet to test, but i'd rather just stay optimistic with this PR, and apply small PRs for any fixes we need in the upstream apps if/when the time comes

src/components/Tooltip/index.tsx
```diff
@@ -7,7 +7,7 @@ const TooltipEl = styled.span`
   position: relative;
 
   .icon {
-    cursor: pointer;
+    cursor: help;
```
Schwartz10, line 10: @navFooh this is the only real change i made besides more cleaning with enums

src/services/WalletProvider/metamaskUtils/state.ts
navFooh: The reason I don't like the desctruring here is that the names (`string`, `number`, `bool`, `object`) are similar or exactly the same as their TypeScript type counterparts. I know the compiler doesn't fall over this, because they're not _used_ as types here, but still it feels like bad practise to me.

Concerning the optimization, my understanding is that it doesn't matter here. It still includes the main 'prop-types' lib file which exports all of the other properties. The destructuring only allows us to access a subset after including them all.

It would actually make a difference if we could include the subset as follows:
```
import bool from 'prop-types/bool'
import string from 'prop-types/string'
```

But as far as I know, `prop-types` doesn't support that. Also for this lib it's probably such a micro optimization that it doesn't matter.

src/components/theme.ts
navFooh: haha yes! 🙌

navFooh:
> **I left you the honors of `npm uninstall styled-system` :D** - let's get that out of the package.json and package-lock

Aii, actually.. I see that it's still used in the `Icons`. And now I think of it, there is another improvement I wanted to make for the `Button/V2` (to use the `base-css` style for the buttons instead of the `styled-component`).

I'd opt to squash and merge this first and then do a final cleanup.

src/components/CopyText/index.tsx
navFooh: Ah I see, yeah with the old behaviour the `Copy` / `Copied` text would basically take up space and be clickable while it had `opacity: 0`. This could cause weird behaviour if you'd want to right-align the copy icon in an element or have other buttons right next to it. I'll have a quick think if we can kind of have the best of both worlds and keep it visible after it pops up.

Or do you also prefer to have the inivisble text as a hit-area before it shows?

navFooh: I've updated the behaviour. Now the `Copy` / `Copied` text does stay visible when hovering over it, but you do need to hover over the icon to show it in the first place.

PR #587, opened 2022-10-16
Update GraphQL types

navFooh:


src/components/HistoryTables/MessageHistory/History/MessagePendingRow.tsx
```diff
@@ -34,7 +39,7 @@ export default function PendingMessageHistoryRow(
           {truncateAddress(message.cid)}
         </SmartLink>
       </td>
-      {props.inspectingAddress && <td>{methodName}</td>}
+      <td>{methodName}</td>
```
Schwartz10, line 42: I'm fine taking this out, but what prompted this change? Why did it happen in this PR? Just curious

src/components/HistoryTables/MessageHistory/History/MessageConfirmedRow.tsx
```diff
@@ -45,14 +48,12 @@ export default function MessageHistoryRow(props: MessageHistoryRowProps) {
...
       </td>
-      {props.inspectingAddress && (
-        <td>
-          <Badge
-            color={toAddressIsInspecting ? 'green' : 'yellow'}
-            text={toAddressIsInspecting ? 'IN' : 'OUT'}
-          />
-        </td>
-      )}
+      <td>
```
Schwartz10, line 51: Fine taking this out, but what prompted this change? Why did it occur in this PR?

src/components/HistoryTables/detail.tsx
```diff
@@ -150,10 +147,10 @@ export const MessageDetailBase = ({
 
   const confirmationCount = useMemo<number>(
     () =>
-      chainHeadSubscription.data?.chainHead.height && !!message?.height
-        ? chainHeadSubscription.data.chainHead.height - Number(message.height)
+      chainHeadSubscription.data?.chainHead.height && message.height
+        ? chainHeadSubscription.data.chainHead.height - message.height
         : 0,
-    [message?.height, chainHeadSubscription.data?.chainHead.height]
+    [message.height, chainHeadSubscription.data?.chainHead.height]
```
Schwartz10, line 153: Are you sure removing the `?` from this doesn't cause an error on initial load?

Schwartz10 (approved):

src/components/HistoryTables/detail.tsx
navFooh: Message is passed as a mandatory prop in `MessageDetailBase`, so it should never render when `message` is null or undefined. In the return statement we were also assuming message to be defined, without the optional chaining. Seems to work fine.

src/components/HistoryTables/MessageHistory/History/MessageConfirmedRow.tsx
navFooh: I think I accidentally came across it :) either way the conditional doesn't seem to be necessary, especially now `inspectingAddress` is a reqiured prop. Also if it _would_ be conditional, we `td` should still be rendered to prevent a column shift

src/components/HistoryTables/MessageHistory/History/MessagePendingRow.tsx
navFooh: Same reason as the other occurence: not necessary to check for `props.inspectingAddress` and we don't want to render the `td` conditionally to prevent content shifting to the wrong column.

PR #590, opened 2022-10-20
Move constant urls to environment

navFooh:


src/components/ErrorView/index.jsx
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import { ButtonV2 } from '../Button/V2'
 import { ButtonRow, Dialog, ErrorBox } from '../Layout'
 import { SmartLink } from '../SmartLink'
-import { GLIF_DISCORD, GLIF_TWITTER } from '../../constants'
+import { useEnvironment } from '../../services'
```
Schwartz10, line 5: Should we path directly into this specific service so we don't "pull all of them in"?

navFooh: Will update some of these in a separate PR

PR #591, opened 2022-10-20
Add ability to import hex pk

Schwartz10:
~~@navFooh this requires publishing of https://github.com/glifio/modules/pull/241~~


src/utils/potentialDupMsg/index.test.ts
```diff
@@ -40,7 +40,8 @@ describe('potentialDupMsg', () => {
       from: WALLET_ADDRESS,
       value: '0',
       method: 0,
-      nonce: 0
+      nonce: 0,
+      params: ''
```
Schwartz10, line 44: This broke the tests since some of the Message types changed, just fixed it in this PR

navFooh (approved):

PR #598, opened 2022-10-25
Update filecoin-address

navFooh:


src/utils/isAddress/index.ts
```diff
@@ -0,0 +1,11 @@
+import { validateAddressString } from '@glif/filecoin-address'
```
Schwartz10, line 1: eventually seems like this file could live in the filecoin-address package, but fine to leave here for now

src/components/InputV2/Select.tsx
```diff
@@ -95,7 +95,7 @@ export const Select = ({
           )}
           {options.map(option => (
             <option key={option} value={option}>
-              {address ? truncateAddress(option) : option}
+              {address ? truncateString(option) : option}
```
Schwartz10, line 98: doesn't need to change in this PR, but the Select InputV2 becomes oddly hard to use when the truncation happens inside of it

src/components/LabeledText/AddressLink.tsx
```diff
@@ -3,7 +3,7 @@ import { useMemo } from 'react'
 import { useAddressQuery } from '../../generated/graphql'
 import { useEnvironment } from '../../services/EnvironmentProvider'
 
-import truncateAddress from '../../utils/truncateAddress'
+import { truncateString } from '../../utils/truncateString'
 import {
```
Schwartz10, line 7: i dont think this will break, although we might want to `skip` the useAddressQuery in here if address is not a valid fil address (dont fetch addresses from the graph if you pass an ethereum address?)

Depends how we want to handle it...

src/services/UpdateBalance/index.tsx
```diff
@@ -20,8 +20,7 @@ const fetcher = async (
 ): Promise<FilecoinNumber | null> => {
   if (!apiAddress || !actorAddress) return null
 
-  if (!validateAddressString(actorAddress))
-    throw new Error('Invalid actor address')
+  if (!isFilAddress(actorAddress)) throw new Error('Invalid actor address')
```
Schwartz10, line 23: idea: what if we used the eth json rpc if balance poller receives an eth address? might be useful for building eth apps?

Schwartz10 (approved):
this PR looks fine, there are some subtle things that seem like dead-ends to me in terms of handling multiple address types, but that requires a larger discussion around how we want to handle things

Also, `useGetMessage` hook - where is that being used? In the transaction replacement?

src/utils/isAddress/index.ts
navFooh: possibly.. not sure if we should clutter that repo with more logic relating to other address types

PR #604, opened 2022-10-27
Fix issue w StateReadState eth addr

Schwartz10:


src/components/InputV2/Address.tsx
```diff
@@ -34,6 +43,28 @@ export const AddressInput = ({
...
+  )
+
+  // Communicate delegated address to parent component
+  useEffect(
+    () => setDelegated(delegatedFromEth),
+    [setDelegated, delegatedFromEth]
+  )
+
+  // Store any converted value for display
+  const converted = delegatedFromEth || ethFromDelegated
```
Schwartz10: Here i think we should change the text based on which way we're going:
```
let suffix = ''

if (delegatedFromEth) suffix = "Converts into ${delegatedFromEth}"
else if (ethFromDelegated) suffix = "Represented by ${ethFromDelegated}"
```
I'm not sure the difficulty on this, but ideally, `delegatedFromEth` and `ethFromDelegated` would be `AddressLink` to the explorer

navFooh: Done!

navFooh: Ooh wait I didn't do the AddressLink

navFooh: Done!

PR #605, opened 2022-10-27
Add ability to swap network in multiple ways

Schwartz10:


src/utils/urlParams/index.ts
```diff
@@ -118,3 +119,38 @@ export const resetWallet = () => {
...
+  } else {
+    url =
+      network === Network.MAINNET
+        ? removeQueryParam(asPath, 'network')
+        : appendQueryParams(asPath, {
+            network
+          })
+  }
+
+  return url
```
navFooh: Looks good to me. Just to be nitpicking I'd change one thing in the code style. Don't create an initial empty `url` that you return in the end, but return directly in the `if` statement. Then there's also no need for an explicit `else`. Makes it a bit more concise:

```javascript
  if (networkInPath) {
    // either get rid of the network name if switching to mainnet, or replace the network name with the new network
    const endPath = asPath.split('/').slice(2)
    return network === Network.MAINNET
        ? ['/', ...endPath].join('/')
        : ['/', network, ...endPath].join('/')
  }
  
  return network === Network.MAINNET
    ? removeQueryParam(asPath, 'network')
    : appendQueryParams(asPath, { network })
```

PR #606, opened 2022-11-01
Fix navigate function when network is encoded as path param

Schwartz10:


src/utils/urlParams/index.ts
```diff
@@ -96,14 +96,33 @@ interface NavigateOptions {
...
 export function navigate(
   router: NextRouter,
   { pageUrl, params, retainParams }: NavigateOptions
 ): void {
   const query = router?.query
   let updatedUrl = pageUrl
 
-  if (query) {
+  if (query && isNetworkInPath(router.asPath)) {
+    updatedUrl += `/${query.network}`
```
navFooh, line 117: I'm a bit confused here. Isn't the network always the first part of the path? E.g. `/calibrationnet/message` and not `/message/calibrationnet`. Also, if `updatedUrl` ends with a `/`, there could be a double slash in the url, I think we should prevent this by checking `updatedUrl.endsWith('/')`

src/utils/urlParams/index.ts
```diff
@@ -96,14 +96,33 @@ interface NavigateOptions {
...
+
+    if (retainParams) {
+      // if the network is already encoded in the url's path, we don't want to add it as a query param
```
navFooh, line 120: This could be an issue with other dynamic route parameters, like `address` and `txID` as well, but not sure how to tackle that right now. Not sure if it will be an issue since I don't think any links in the explorer have the need to retain parameters.

navFooh (changes requested):
I feel that any changes we make to the logic here also needs to be carried over to `SmartLink`. It's getting a bit messy with the logic of query params and I'm wondering if we can deprecate some of this.

navFooh:
@Schwartz10 I do think we've been going at it wrong with the network name path parameter. Instead of `/wallabynet/tx/...` I think it should be `/tx/.../wallabynet`.

The reason for this is that routes will be way easier to maintain like this. For now it's not super important because the explorer only has two pages. But imagine if we have the mono-app with a ton of pages and we have to make aliases for everything to support prefixing every path with the network name.

Next.js has a thing called [catch all routes](https://nextjs.org/docs/routing/dynamic-routes#catch-all-routes) which would enable us to get both a `txID` and network name from the end of the path, where network name can be optional.

That said, I think even better would be:
- `mainnet.glif.io/tx/...` | `glif.io/tx/...`
- `wallabynet.glif.io/tx/...`
- `calibrationnet.glif.io/tx/...`

But that will only work when we have the mono app setup.

Schwartz10:
Hey @navFooh like we talked about briefly in discord - the explorer URL API spec we're following is https://eips.ethereum.org/EIPS/eip-3091 since metamask follows it - which means we can't encode the network at the end of the path

I think it looks a bit cleaner to separate the sites per network via subdomains, but since testnet documentation contains a lot of backlinks to glif, it makes sense to keep this on one domain to bump our SEO traffic

src/utils/urlParams/index.ts
Schwartz10: oh right... hmm will think about this a bit more too

navFooh:
> Hey @navFooh like we talked about briefly in discord - the explorer URL API spec we're following is https://eips.ethereum.org/EIPS/eip-3091 since metamask follows it - which means we can't encode the network at the end of the path
> 
> I think it looks a bit cleaner to separate the sites per network via subdomains, but since testnet documentation contains a lot of backlinks to glif, it makes sense to keep this on one domain to bump our SEO traffic

I think we might be able to do this with [optional catch-all routes](https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes) as well: `[[network]]/address/[address]`. That way we don't have to redefine all our pages with and without the network in the path. However this doesn't work in Next.js 12, so we have to upgrade to 13 first.

Yeah, makes sense to follow the spec. In that case there are still three things we should look at:
- [This comment](https://github.com/glifio/react-components/pull/606#discussion_r1011707150) about prepending instead of appending the network
- Updating the logic in the SmartLink component
- excluding other path parameters from the `retainParams` functionality. I think we can easily check for each params presence in the router path (which doesn't include query params)

PR #613, opened 2022-11-14
initial plumbing for inspecting fevm txs

Schwartz10:
use in conjuction with https://github.com/glifio/modules/pull/252


src/components/HistoryTables/MessageHistory/Detail/MethodName.tsx
```diff
@@ -0,0 +1,18 @@
+import { Address } from '../../../../generated/graphql'
+import { Badge, Line } from '../../../Layout'
+import { useMethodName } from '../hooks/useMethodName'
+
+export function MethodName({ address, params, methodNum }: MethodNameProps) {
```
Schwartz10: Separating out this component fixed an issue with useAbi not updating after setting undefined as an initial value

src/components/HistoryTables/MessageHistory/hooks/useMethodName.ts
```diff
@@ -18,6 +24,13 @@ export const useMethodName = (
     skip: !address
   })
 
+  const [abi] = useAbi(address?.robust)
+  if (isDelegatedAddress(address?.robust) && !!abi) {
+    return getFEVMMethodName(params, abi)
+  } else if (isDelegatedAddress(address?.robust)) {
+    return 'Unknown'
+  }
+
```
Schwartz10: Something in here errors now on the transaction history view but not the error detail view, not sure if the `abi` has something to do with this...

src/components/Layout/DataTypes.tsx
```diff
@@ -63,7 +63,7 @@ export const DataTypeLines = ({
       )
 
     default:
-      throw new Error(`Unexpected DataType: ${JSON.stringify(dataType)}`)
+      return <Line label={dataType.Name}>{dataType.Value.toString()}</Line>
```
Schwartz10: This is a hack until we figure out the data types question

src/components/InputV2/File.tsx
```diff
@@ -0,0 +1,12 @@
+import { ChangeEventHandler } from 'react'
+
+export const FileInput = ({ accept, handleFileUpload }: FileInputProps) => {
+  return <input type='file' onChange={handleFileUpload} accept={accept} />
```
Schwartz10: placeholder for a file input

src/utils/useAbi.ts
```diff
@@ -0,0 +1,42 @@
+import { JsonFragment } from '@ethersproject/abi'
+import { Fragment } from 'ethers/lib/utils'
+import { useState } from 'react'
+
+export type ABI = ReadonlyArray<Fragment | JsonFragment | string>
+
+export const useAbi = (actorAddr: string) => {
```
Schwartz10: theres an annoying bug where - when you set your contract in local storage, you have to refresh the page in order for it to take effect. this is because the upload input renders its own `useAbi` hook, and doesn't share the same state with its sibling components that also render the `useAbi` hook 

there are two options i see:

1. make a context wrapper so all useAbi hooks share the same state
2. change the prop structure as to only use the useAbi hook once

I'm thinking 1 probably makes more sense?

Schwartz10:
@navFooh there's a bug in here around the method name on the transaction history just FYI

src/components/HistoryTables/MessageHistory/Detail/MethodName.tsx
navFooh: I think this might have had something to do with SSR. Moving it into a piece of code that only renders in the browser makes sure window is defined and the right value will be set. The new useLocalStorage hook makes sure the value resaturated when going from SSR to the browser.

src/components/HistoryTables/MessageHistory/hooks/useMethodName.ts
navFooh: Might also have to do with SSR

PR #614, opened 2022-11-15
useAbi hook improvements

Schwartz10:


Schwartz10 (commented):
@navFooh i think theres a more efficient way to do the useLocalStorage hook internal logic.. but this should work fine for now and my hope is we remove this all together with verified bytecode uploads

PR #620, opened 2022-11-18
A bunch of small, hacky improvements for the hackathon

Schwartz10:


src/components/Layout/LinesReturn.tsx
```diff
@@ -103,6 +103,8 @@ export const LinesReturn = ({
     <DataTypeLines label='Return' dataType={describedReturnVal} />
   ) : isDelegated && !abi ? (
     <Line label='Return (upload abi to decode)'>{returnVal}</Line>
+  ) : isDelegated ? (
+    <NullishLine label='Return' />
```
navFooh: @Schwartz10 I'm not sure why we should add the `NullishLine` here? If the actor is delegated and we do have the ABI, then it should either parse to a `describedReturnVal` or we should show that it failed to decode with the raw `returnVal`. Showing a `NullishLine` just makes it seem like there is not return value altogether.

src/components/Layout/LinesParams.tsx
```diff
@@ -99,14 +99,10 @@ export const LinesParams = ({ address, method, params }: LinesParamsProps) => {
...
-      label={`Params (${
-        decodeParamsError ? 'failed to decode' : 'decoding...'
-      })`}
-    >
-      {params}
-    </Line>
+  ) : params && decodeParamsError ? (
+    <Line label='Params (failed to decode)'>{params}</Line>
+  ) : params && !isDelegated ? (
+    <Line label='Params (decoding...)'>{params}</Line>
```
navFooh: I think this seems to prevent a `(decoding...)` state for the params when the actor is delegated, which might make sense because parameter decoding is synchronous. But I think it also causes to now just show the `NullishLine` if the parameters for a smart contract tx fail to decode for some reason.

navFooh (approved):

PR #621, opened 2022-11-19
[fix] - changed node link to the right one

mur-me:


mur-me:
@Schwartz10 please, take a look :)

